PR #1, opened 2015-03-19
Adopt nullability annotations for cleaner Swift headers

cxa:
Nullability annotations is more friendly to Swift. Not break compatibility for Xcode less than 6.3.

### DIFF
--- a/PINCache/nullability.h
+++ b/PINCache/nullability.h
@@ -1,0 +1,24 @@
+//
+//  nullability.h
+//  PINCache
+//
+//  Created by CHEN Xian’an on 3/19/15.
+//  Copyright (c) 2015 Tumblr. All rights reserved.
+//
+
+#ifndef PINCache_nullability_h
+#define PINCache_nullability_h
+
+#if !__has_feature(nullability)
+#define NS_ASSUME_NONNULL_BEGIN
+#define NS_ASSUME_NONNULL_END
+#define nullable
+#define nonnull
+#define null_unspecified
+#define null_resettable
+#define __nullable
+#define __nonnull
+#define __null_unspecified
+#endif
+
+#endif
--- a/PINCache/Nullability.h
+++ b/PINCache/Nullability.h
@@ -1,9 +1,6 @@
-//
-//  nullability.h
-//  PINCache
-//
-//  Created by CHEN Xian’an on 3/19/15.
-//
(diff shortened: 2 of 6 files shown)


garrettmoon:
Awesome! Thanks for this. I'll look at this soon!

garrettmoon:
Looks like in PINCache.h
    - (void)objectForKey:(NSString *)key block:(PINCacheObjectBlock)block;
    - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key block:(PINCacheObjectBlock)block;
    - (void)removeObjectForKey:(NSString *)key block:(PINCacheObjectBlock)block;
    - (void)trimToDate:(NSDate *)date block:(PINCacheBlock)block;
    - (void)removeAllObjects:(PINCacheBlock)block;

should all have nullable blocks?

cxa:
Oh, nice catches. Except this one `- (void)objectForKey:(NSString *)key block:(PINCacheObjectBlock)block;`, the implementation begins with `if (!key || !block) return;`.

I'll push a new commit if we can confirm block in `- (void)objectForKey:(NSString *)key block:(PINCacheObjectBlock)block;` should not be nullable.

garrettmoon:
You are correct, objectForKey: block: should *not* be nullable.

PINCache/nullability.h
```diff
@@ -0,0 +1,24 @@
+//
+//  nullability.h
+//  PINCache
+//
+//  Created by CHEN Xian’an on 3/19/15.
+//  Copyright (c) 2015 Tumblr. All rights reserved.
```
garrettmoon: Can you either remove this copyright?

garrettmoon:
Also, it looks like there may be missing ones in both PINMemoryCache and PINDiskCache?

cxa:
OK, I add nullable-s by the comments that allow nil, if not specific, I spot the implementations.

One exception, according to the implementation, those should be failable initializer, be it depends on the `name` null or not. If we don't allow name to be null, the initializer should be success, so I don't annotate those initializers.

`- (instancetype)initWithName:(NSString *)name;` and
`- (instancetype)initWithName:(NSString *)name rootPath:(NSString *)rootPath NS_DESIGNATED_INITIALIZER;`

garrettmoon:
Thanks looks good to me!

PR #3, opened 2015-04-18
Adding unit tests

gmcnaughton:
Added a handful of unit tests:
* `setObject`: setting the same key twice overwrites the original value
* `objectForKey`: getting a key that does not exist returns nil
* `removeAllObjects`: removes all objects from the cache
* `initWithName`: caches with the same name share the same contents

### DIFF
--- a/tests/PINCacheTests/PINCacheTests.m
+++ b/tests/PINCacheTests/PINCacheTests.m
@@ -95,6 +95,28 @@ NSTimeInterval PINCacheTestBlockTimeout = 5.0;
     XCTAssertNotNil(image, @"object was not set");
 }
 
+- (void)testObjectSetWithDuplicateKey
+{
+    NSString *key = @"key";
+    NSString *value1 = @"value1";
+    NSString *value2 = @"value2";
+    __block NSString *cachedValue = nil;
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+    
+    [self.cache setObject:value1 forKey:key];
+    [self.cache setObject:value2 forKey:key];
+    
+    [self.cache objectForKey:key block:^(PINCache *cache, NSString *key, id object) {
+        cachedValue = (NSString *)object;
+        dispatch_semaphore_signal(semaphore);
+    }];
+    
+    dispatch_semaphore_wait(semaphore, [self timeout]);
+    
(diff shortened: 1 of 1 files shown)


tests/PINCacheTests/PINCacheTests.m
```diff
@@ -95,6 +95,28 @@ - (void)testObjectSet
     XCTAssertNotNil(image, @"object was not set");
 }
 
+- (void)testObjectSetWithDuplicateKey
+{
+    NSString *key = @"key";
+    NSString *value1 = @"value1";
+    NSString *value2 = @"value2";
+    __block NSString *value = nil;
```
garrettmoon: Mind renaming this to something indicating it's the returned value? Maybe cachedValue?

tests/PINCacheTests/PINCacheTests.m
```diff
@@ -131,6 +172,29 @@ - (void)testObjectRemove
...
+    }];
+    
+    dispatch_semaphore_wait(semaphore, [self timeout]);
+    
+    id object1 = [self.cache objectForKey:key1];
+    id object2 = [self.cache objectForKey:key2];
+    
+    XCTAssertNil(object1, @"not all objects were removed");
+    XCTAssertNil(object2, @"not all objects were removed");
+    XCTAssertTrue(self.cache.memoryCache.totalCost == 0, @"memory cache was not 0 after removing all objects");
```
garrettmoon: Mind adding this too:
    XCTAssertTrue(self.cache.diskByteCount == 0, @"disk cache byte count was not 0 after removing all objects");

garrettmoon:
Thank you for writing these! If you don't mind addressing a couple of comments I'll get this merged in!

tests/PINCacheTests/PINCacheTests.m
gmcnaughton: Yep, renamed to cachedValue!

tests/PINCacheTests/PINCacheTests.m
gmcnaughton: Yep of course - added this assert!

gmcnaughton:
@garrettmoon Thanks!  I made the changes you suggested and force-pushed the branch.

PR #11, opened 2015-07-28
Set background task to invalid if expiration handler runs.

garrettmoon:


garrettmoon:
Apple's example code shows this:
https://developer.apple.com/library/ios/documentation/iPhone/Conceptual/iPhoneOSProgrammingGuide/BackgroundExecution/BackgroundExecution.html

garrettmoon:
Hmm, this solution isn't thread safe :/ I'll have to think about this some more.

garrettmoon:
Actually, this doesn't even make sense: #13

Issue #13, opened 2015-07-28
Invalid task ids used for expiration handlers.


jaredsinclair:
The macro for beginning a background task looks like this:

```
UIBackgroundTaskIdentifier taskID = UIBackgroundTaskInvalid;
taskID = [[UIApplication sharedApplication] beginBackgroundTaskWithExpirationHandler:^{
    [[UIApplication sharedApplication] endBackgroundTask:taskID]; 
}];
```

`UIBackgroundTaskIdentifier` is a value type, so when the block is created it captures `taskID` when it equals `UIBackgroundTaskInvalid`. It never captures the actual task identifier, so if the expiration handler pops the actual task isn't properly cancelled.

garrettmoon:
Yup.

garrettmoon:
@jaredsinclair Let me know what you think of #14

jaredsinclair:
# lg™

PR #14, opened 2015-07-28
Invalid task ids used for expiration handlers. #13

garrettmoon:
This should fix two issues:
- Because taskID was being captured before being set, the expiration handler had an invalid taskID.
- If the expiration handler was called, and then PINCacheEndBackgroundTask was called, it would try to end a task that had already ended.

The second case is less of an issue because I don't see it causing any problems. I've used an atomic getter / setter for taskID which should make it far less likely for there to be a race condition where endBackgroundTask is called twice with the same taskID. However it's still possible. Since this just seems to generate warnings, this seems better than the overhead of introducing a lock of some sort.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -12,36 +12,13 @@
 [[NSString stringWithUTF8String:__FILE__] lastPathComponent], \
 __LINE__, [error localizedDescription]); }
 
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0 && !defined(PIN_APP_EXTENSIONS)
-#define PINCacheStartBackgroundTask() \
-PINBackgroundTask *task = [[PINBackgroundTask alloc] init]; \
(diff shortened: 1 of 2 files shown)


jparise:
If we're going to introduce any intermediate object to hold the task identifier, maybe just go all the way and wrap the full background task interaction (instead of using a macro):

```objc
PINBackgroundTask *task = [[PINBackgroundTask alloc] initWithExpirationHandler:^{ ... }];
// ...
[task end];

**Comment by jaredsinclair** [comment]
**lgtm.** It might be nice to add a convenience class method like:

```
PINBackgroundTask *task = [PINBackgroundTask start];
```

to keep the one-liner aesthetic of the original macro.

jparise:
I'd prefer the `PINBackgroundTask *task = [PINBackgroundTask start];` version, too.

garrettmoon:
Sounds good, I'll update when I get access to my computer :)

PR #21, opened 2015-07-31
Fix tests related to disk cache and conflicting names

garrettmoon:
- Change setup / tear down to use UUID as name. It is no longer safe to use two disk caches with the same name.
- Remove invalid test.
- Fix disk cache directories being created asynchronously.


PINCache/PINDiskCache.m
```diff
@@ -89,14 +89,10 @@ - (instancetype)initWithName:(NSString *)name rootPath:(NSString *)rootPath
...
         
-        __weak PINDiskCache *weakSelf = self;
-        dispatch_async(_asyncQueue, ^{
-            PINDiskCache *strongSelf = weakSelf;
-            [strongSelf lock];
-                [strongSelf createCacheDirectory];
-                [strongSelf initializeDiskProperties];
-            [strongSelf unlock];
-        });
+        [self lock];
```
jparise: Do you still need to lock this section while we're still in the initializer?  The lock (semaphore) is per-instance.

garrettmoon: Nope.

PR #22, opened 2015-08-01
Don't need to lock in initialization

garrettmoon:
This was meant to go into a previous commit (thanks @jparise).


jparise:
:+1:

PR #23, opened 2015-08-04
fileURL in PINDiskCacheObjectBlock should be nullable

xhacker:


garrettmoon:
Indeed, thank you! Looks like Travis CI is taking a while with the build, I'll merge this in once it's finished though.

PR #25, opened 2015-08-20
Fix Xcode 7 warnings

garrettmoon:


jparise:
I think we should consider adding:

```objc
- (instancetype)init NS_UNAVAILABLE;
```

... to the class extension interface, too.

tapz:
2.0 pods does not seem to have this fix. At least Xcode 7 gives the error.

garrettmoon:
It won't be in cocoa pods until there is a new release.

garrettmoon:
Will try and get one out soon.

garrettmoon:
2.1 should include this.

Issue #41, opened 2015-11-19
#36 is breaking the disk cache of PINCache


mRs-:
https://github.com/pinterest/PINCache/pull/36 is breaking the disk cache of PINCache

more explanation is directly in the pull request.

garrettmoon:
Should be fixed by #42

PR #36, opened 2015-10-29
Bugfix/28

ccerinojr:
Fixing deprecation warnings for:
CFURLCreateStringByAddingPercentEscapes
CFURLCreateStringByReplacingPercentEscapesUsingEncoding

This addresses issue #28

PR #42, opened 2015-11-19
[BUGFIX] Critical! String Encoding Bugfix

mRs-:
As mentioned in https://github.com/pinterest/PINCache/issues/41 I created a bugfix for https://github.com/pinterest/PINCache/pull/36

Maybe this should be unit tested, because it broke the complete DiskCache :sob:. I prepared a test for this but cannot access the private method!

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -143,7 +143,7 @@ NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
 {
     if (![string length])
         return @"";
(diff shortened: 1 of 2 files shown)


garrettmoon:
You can access a private method in tests by creating an interface for it in the tests file.

mRs-:
done :+1:

garrettmoon:
Thanks for fixing this! I would have thought our disk caching tests would have caught this, but they weren't checking for URLs :(

Thank you!

Issue #50, opened 2015-11-25
Long keys don't work for disk cache due to filename length limit


shwu:
The symptom we are seeing is that keys above a certain length cause the cache to either not save or miss when accessed.  After a quick empirical experiment it seems like the maximum key length is 255 characters (UNIX filename limit) and since special characters are escaped, their presence lowers the maximum limit.

We're using URLs as keys, some of which can be very long.

garrettmoon:
Hi @shwu, can you be a bit more specific? Is this specifically the disk cache?

erichoracek:
I've encountered this issue as well when caching Mapbox static map images with very long URLs. Applying a SHA1 to the key solved the issue for me. However, it would probably make more sense to fix this at the framework level.

shwu:
@garrettmoon Yep, this is specifically the disk cache.  When trying to write to disk with a long key (over 255 characters after escaping) the object will not be saved.  Our workaround was to hash the URL, as @erichoracek suggested, but again it seems like a framework-level fix is appropriate in this case.

shwu:
Awesome!

PR #52, opened 2015-12-03
If filenames are too long, sha1 them and hope for no collisions

garrettmoon:
Should address #50

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -169,13 +169,17 @@ static NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
     size_t length = strlen(fileSystemRepresentation);
     
     if (length > NAME_MAX) {
-        //encoded string name will be too long for file representation, md5 and hope for no collisions.
+        //encoded string name will be too long for file representation, sha1 and hope for no collisions.
         unsigned char sha1Buffer[CC_SHA1_DIGEST_LENGTH];
         CC_SHA1(fileSystemRepresentation, (CC_LONG)length, sha1Buffer);
-        encodedString = @"";
+        
+        static char const *hexChars = "0123456789ABCDEF";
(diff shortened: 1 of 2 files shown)


PINCache/PINDiskCache.m
```diff
@@ -143,7 +146,24 @@ - (NSString *)encodedString:(NSString *)string
...
+    const char *fileSystemRepresentation = [encodedString fileSystemRepresentation];
+    size_t length = strlen(fileSystemRepresentation);
+    
+    if (length > NAME_MAX) {
+        //encoded string name will be too long for file representation, md5 and hope for no collisions.
+        unsigned char sha1Buffer[CC_SHA1_DIGEST_LENGTH];
+        CC_SHA1(fileSystemRepresentation, (CC_LONG)length, sha1Buffer);
+        encodedString = @"";
+        for (NSUInteger idx = 0; idx < CC_SHA1_DIGEST_LENGTH; idx++) {
+            encodedString = [encodedString stringByAppendingFormat:@"%02x", sha1Buffer[idx]];
```
garrettmoon: There is probably a more efficient way of doing this…

jparise:
MD5 is generally a faster hash algorithm than SHA1. That speed difference probably isn't relevant here given such short input strings. Is there a specific reason you chose SHA1 though?

PINCache/PINDiskCache.m
jparise: If you're going to build up a string like this, you'd be better using an `NSMutableString` or, even better, writing into a C character buffer on the stack (because you know the exact buffer length you need) and then converting that to an `NSString` for the return value.

garrettmoon:
The intention was that SHA1 is much less likely to have a collision. Specifically because I'm not handling collisions in any way and having one would result in very bad behavior. That's the thought anyhow, I'd love it if there was a better solution.

garrettmoon:
Back to the drawing board on this one, the filenames need to be able to be transformed back into keys :(

PR #53, opened 2015-12-09
moved from PIN_APP_EXTENSIONS to NS_EXTENSION_UNAVAILABLE_IOS

mRs-:
Needs code review + testing, but must be okay IMHO


garrettmoon:
As mentioned [here](https://github.com/pinterest/PINRemoteImage/issues/93), I don't think this will work :(

PR #55, opened 2015-12-18
Fix decodedString / encodedString methods in PINDiskCache.m to suppor…

jlukanta:
…t SDKs older than 7.0.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -146,15 +146,18 @@ NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
     }
     
     if ([string respondsToSelector:@selector(stringByAddingPercentEncodingWithAllowedCharacters:)]) {
-        return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet characterSetWithCharactersInString:@".:/"]];
+        return [string stringByAddingPercentEncodingWithAllowedCharacters:[[NSCharacterSet characterSetWithCharactersInString:@".:/"] invertedSet]];
     }
     else {
         CFStringRef static const charsToEscape = CFSTR(".:/");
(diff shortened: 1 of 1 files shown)


jlukanta:
iOS 6.0 devices would crash without this fix

PINCache/PINDiskCache.m
```diff
@@ -141,16 +141,40 @@ - (NSString *)keyForEncodedFileURL:(NSURL *)url
...
 - (NSString *)encodedString:(NSString *)string
 {
-    if (![string length])
+    if (![string length]) {
         return @"";
-   return [string stringByAddingPercentEncodingWithAllowedCharacters:[[NSCharacterSet characterSetWithCharactersInString:@".:/"] invertedSet]];
+    }
+    
+    if ([string respondsToSelector:@selector(stringByAddingPercentEncodingWithAllowedCharacters:)]) {
+        return [string stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet characterSetWithCharactersInString:@".:/"]];
```
garrettmoon: shouldn't this be the invertedSet?

garrettmoon:
CFURLCreateStringByAddingPercentEscapes is deprecated and thus will cause warnings. Can you think of a way to only add the code path in if the target supports iOS 6?

jlukanta:
Good catch! Fixed!

jlukanta:
In general, I think it's better to suppress the deprecated warning instead of checking for the OS version because it requires no knowledge of when the functions were deprecated in all supported OS-es (iPhone and Mac). It should be less prone to errors, and results in cleaner code (less checks for different types of OS-es).

But I am fine either way, and I can change it if you think the other approach is better.

garrettmoon:
Nope, I didn't notice the pragmas. I like your way better! Thanks!

PR #56, opened 2015-12-29
Fix static declarations

hoppenichu:
As titled.


jparise:
Thanks! I doubt anyone was relying on these symbols being link-able, especially without any formal `extern` declarations in the headers.

hoppenichu:
🍻

PR #57, opened 2015-12-29
Add weak cache to keep references to retained objects

lukescholefield:
The map table ensures the cache has a reference to an object still in memory, even if it has been evicted from the dictionary.

### DIFF
--- a/PINCache/PINMemoryCache.m
+++ b/PINCache/PINMemoryCache.m
@@ -59,7 +59,9 @@ static NSString * const PINMemoryCachePrefix = @"com.pinterest.PINMemoryCache";
         _concurrentQueue = dispatch_queue_create([queueName UTF8String], DISPATCH_QUEUE_CONCURRENT);
 
         _dictionary = [[NSMutableDictionary alloc] init];
-        _weakMapTable = [NSMapTable strongToWeakObjectsMapTable];
+        if ([[NSMapTable class] respondsToSelector:@selector(strongToWeakObjectsMapTable)]) {
+            _weakMapTable = [NSMapTable strongToWeakObjectsMapTable];
+        }
         _dates = [[NSMutableDictionary alloc] init];
         _costs = [[NSMutableDictionary alloc] init];
 
--- a/PINCache/PINMemoryCache.m
+++ b/PINCache/PINMemoryCache.m
@@ -19,6 +19,7 @@ static NSString * const PINMemoryCachePrefix = @"com.pinterest.PINMemoryCache";
 @property (assign, nonatomic) dispatch_semaphore_t lockSemaphore;
 #endif
 @property (strong, nonatomic) NSMutableDictionary *dictionary;
+@property (strong, nonatomic) NSMapTable *weakMapTable;
 @property (strong, nonatomic) NSMutableDictionary *dates;
 @property (strong, nonatomic) NSMutableDictionary *costs;
 @end
@@ -58,6 +59,7 @@ static NSString * const PINMemoryCachePrefix = @"com.pinterest.PINMemoryCache";
         _concurrentQueue = dispatch_queue_create([queueName UTF8String], DISPATCH_QUEUE_CONCURRENT);
 
         _dictionary = [[NSMutableDictionary alloc] init];
+        _weakMapTable = [NSMapTable strongToWeakObjectsMapTable];
         _dates = [[NSMutableDictionary alloc] init];
         _costs = [[NSMutableDictionary alloc] init];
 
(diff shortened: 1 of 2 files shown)


PINCache/PINMemoryCache.m
```diff
@@ -383,18 +385,23 @@ - (void)enumerateObjectsWithBlock:(PINMemoryCacheObjectBlock)block completionBlo
...
+
+        if (!object) {
+            object = [_weakMapTable objectForKey:key];
+        }
     [self unlock];
         
     if (object) {
         [self lock];
+            NSDate *now = [[NSDate alloc] init];
             _dates[key] = now;
```
jparise: Or just `_dates[key] = [[NSDate alloc] init];`

lukescholefield: Updated.

lukescholefield:
@jparise Sorry for the changes this morning, I had intended to open the PR against our branch so my colleagues could check through.  Clearly I missed a dropdown.  Updates have been made.  

If you would like to reach out to me, I can be found using my first name @ the company I work for.

jparise:
The date stuff looks great. I would accept that immediately as a solo PR.

The weak map bits look useful too, but I think we might want to include some unit test coverage for that new behavior.

lukescholefield:
There is also the issue of behavior on `-removeAllObjects`

I have not removed objects from the weak cache for this, since it is called for the app entering background and also memory warning.  Since the objects referenced by the weak cache are retained elsewhere, this won't removing them won't save memory.  

A similar issue exists for `-removeObjectAndExecuteBlocksForKey:`.

The primary issue is that both these methods are exposed, so potentially there could be some confusion about expected behavior.

jparise:
@garrettmoon is probably the best person to offer guidance on those issues.

lukescholefield:
Fixed failing unit tests. Primary issue came from objects remaining in the (new weak) cache when they were not deallocated, which is the desired behavior.  Added (admittedly only one) test coverage for weak cache. 

Very happy to take other feedback/questions.

lukescholefield:
@jparise If you would like me to remove the `NSDate` change from this PR and submit it separately, please let me know.

jparise:
@lukescholefield Yes, if you have a moment to do that, we can get that part merged straight away.

lukescholefield:
Removed commit related to NSDate changes, which is now in a separate PR.  Rebased to latest version of master.  Sorry for the forced pushes, but I am trying to keep this in a conflict free state (and keep a tidy commit history).

garrettmoon:
This is an improvement I've been wanting to make! Thank you for taking the time to improve the project!

One thing that has been stopping me is how we can make this behavior optional (precisely because people will expect removeAllObjects to actually remove all objects). I feel like it's probably the correct default behavior, however, I'm concerned about making it such without a point release.

Can you think of a way to make this optional behavior? One way would be to not init the weakMap unless the cache were configured to use it? Then when we decide to roll out PINCache 3 we can init it by default?

lukescholefield:
I had a few options:

1.  Make weak cache optional/configurable as you suggest.
2.  Implement either force remove methods for flushing the weak cache too or add a parameter to the existing remove methods that controls if the weak cache is cleared (and provide convenience remove methods consistent with the current API).
3.  Implement the weak cache and document the behavior (which is what I have done here, minus the documentation).
4. Modifying `-setObject:forKey:withCost:` to clear remove cached object when `object == nil` to provide a way for the user to flush individual objects out of the cache.

As I mentioned before (and you no doubt know), failing to remove the object from the weak cache doesn't really cost memory (aside from the references), so the only downside to the object sticking around is if the user wants to flush it for a specific reason. In this case I would rather push the burden onto the user not to check the cache, but that is a personal opinion, and would at least require documentation.

So I would still lean towards 3, as it results in the simplest API, but you are entirely valid with your concerns about including it in a point release. Implementing 4 would possibly be useful too.

The other option I had was to introduce it as an entirely entirely new layer of cache in addition to `PINDiskCache` and `PINMemoryCache`. If this is preferred it can easily be done, but I felt this change was more concise and logically consistent (since after all, it is just an additional type of memory cache).

timonus:
Err, when I said `NSHashTable`'s `weakObjectsHashTable` method I *actually* meant `NSMapTable`'s `strongToWeakObjectsMapTable`.

screen shot 2016-01-11 at 4 32 20 pm

I'll update the prior comments to be accurate.

lukescholefield:
@timonus Updated with check in 524440c

This should now be fine back to iOS 5.

garrettmoon:
@lukescholefield I think we should at the very least remove objects from the map table when removeObject is called from the API.

garrettmoon:
Hey @lukescholefield I know you're busy, so no pressure on this, but I'm going to close this until you're ready to update. Ping me when you are and I'll re-open.

PR #59, opened 2016-01-02
Only initialize NSDate when it is needed

lukescholefield:
Further to discussion in https://github.com/pinterest/PINCache/pull/57 this PR cleans up date logic so that we only create the `NSDate` when we actually need to use it.


garrettmoon:
Awesome, thank you!

PR #60, opened 2016-01-05
Fix variable shadowing warnings

### LINKED ISSUE #51: Hidden local variables
There are quite a few local variables that shadow other local variables.
For example, the weakSelf variable in PINCache's objectForKey method.

maicki:
Should fix #51 

For the PINCache's objectForKey method I added a clang compiler pragma to suppress showing up the shadowing warnings. This is only for the weakSelf / strongSelf variables. There should be no other variables be shadowed anymore.

We could certainly go without the clang pragma, we should just come up with good names for all of this shadowed weakSelf / strongSelf variables than :)

### DIFF
--- a/PINCache/PINCache.m
+++ b/PINCache/PINCache.m
@@ -74,6 +74,8 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
 
 #pragma mark - Public Asynchronous Methods -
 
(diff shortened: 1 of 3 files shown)


garrettmoon:
I think the pragma makes sense in this case. Thanks for doing this!

Issue #32, opened 2015-10-12
*forKey: and *forKey:block: methods should check object dates against _ageLimit


benasher44:
Hello! In a perfect world, setting ```ageLimit``` schedules regular trims of the cache, and the the cache is always free of objects that have expired. In reality, you could set an ```ageLimit``` to 5 minutes, add an object to the cache 2.5 minutes later, and then this object will get to live on for 7.5 minutes instead of the 5 minutes maximum set by setting the ```ageLimit``` (because it missed the cache trim at the 5 minute mark, and it won't get cleared again until the 10 minute mark).

To avoid this scenario entirely, methods like ```objectForKey:``` and ```objectForKey:block:``` should only return objects in the cache if they're actually meant to still be in the cache (i.e. return nil if the object has technically expired but hasn't been cleared yet). Another benefit, aside from avoiding the scenario I mentioned earlier, is that this allows the user to reliably write code like this:

```
[self.someCache objectForKey:myObjectKey block:^(PINDiskCache *cache, NSString *key, id <NSCoding> __nullable object, NSURL * __nullable fileURL) {
    if (object) {
        // use the object
    } else {
        // create/fetch a new one of these objects
    }
}];
```

Without this, you first have to get the ```fileURL```, check the ```fileURL``` file's modification time and compare it against the ```ageLimit```, if that all checks out, then call ```objectForKey:block:```. The issue here though is that if you do the ```fileURL``` file modification time checking asynchronously to avoid holding any locks in PINCache, the following ```objectForKey:block:``` call is not guaranteed to return your object as a ```trimToAgeLimitRecursively``` may have been called during that time and cleared your object!

Does this make sense? Thoughts?

garrettmoon:
Hmm, this is an interesting use of the cache. I don't think the current design of PINCache is catered to the idea that the cache *must* enforce TTL. I think it's catered to the other use case where you generally want to clear out old items so they don't take up too much memory, not because you want to make sure you 'create/fetch' new objects.

If you feel strongly about adding the feature, I'd be open to a pull request that allowed the feature to optionally be enabled and that didn't cause a hit to performance when not enabled.

benasher44:
[...]

PR #63, opened 2016-01-23
Added TTL cache capability

benasher44:
In #32, I wanted to add what basically amounts to the ability to make the caches behave like a TTL cache. I've done this with tests, which I think clearly illustrate the behavior I was going for here. I'm pretty open to discussion on this, and I've made sure that this is opt-in, so existing behavior of the caches should be unaffected.

### DIFF
--- a/PINCache/PINDiskCache.h
+++ b/PINCache/PINDiskCache.h
@@ -105,7 +105,7 @@ typedef void (^PINDiskCacheObjectBlock)(PINDiskCache *cache, NSString *key, id <
       the cache will behave as if the object does not exist
  
  */
-@property (assign, getter=isTTLCache) BOOL ttlCache;
+@property (nonatomic, assign, getter=isTTLCache) BOOL ttlCache;
 
 #pragma mark -
 /// @name Event Blocks
--- a/PINCache/PINDiskCache.h
+++ b/PINCache/PINDiskCache.h
@@ -97,6 +97,16 @@ typedef void (^PINDiskCacheObjectBlock)(PINDiskCache *cache, NSString *key, id <
  */
 @property (assign) NSTimeInterval ageLimit;
 
+/**
+ If ttlCache is YES, the cache behaves like a ttlCache. This means that once an object enters the
+ cache, it only lives as long as self.ageLimit. This has the following implications:
+    - Accessing an object in the cache does not extend that object's lifetime in the cache
+    - When attempting to access an object in the cache that has lived longer than self.ageLimit,
+      the cache will behave as if the object does not exist
+ 
+ */
+@property (assign, getter=isTTLCache) BOOL ttlCache;
+
 #pragma mark -
 /// @name Event Blocks
 
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -618,7 +618,9 @@ static NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
         fileURL = [self encodedFileURLForKey:key];
         object = nil;
         
-        if ([[NSFileManager defaultManager] fileExistsAtPath:[fileURL path]]) {
+        if ([[NSFileManager defaultManager] fileExistsAtPath:[fileURL path]] &&
+            // If the cache should behave like a TTL cache, then only fetch the object if there's a valid ageLimit and  the object is still alive
+            (!self->_ttlCache || self->_ageLimit <= 0 || fabs([[_dates objectForKey:key] timeIntervalSinceDate:now]) < self->_ageLimit)) {
             @try {
                 object = [NSKeyedUnarchiver unarchiveObjectWithFile:[fileURL path]];
             }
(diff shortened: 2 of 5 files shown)


garrettmoon:
Neat! I'll try and look at this soon (next week by the latest). My greatest concern about this is making sure there are no performance impacts if this new feature is disable.

benasher44:
@garrettmoon cool thanks! I think it should be alright performance-wise. In all cases, I check `!self->_ttlCache` before starting the TTL cache behavior, so performance impact should be minimal

PINCache/PINDiskCache.h
```diff
@@ -97,6 +97,16 @@ typedef void (^PINDiskCacheObjectBlock)(PINDiskCache *cache, NSString *key, id <
...
 
+/**
+ If ttlCache is YES, the cache behaves like a ttlCache. This means that once an object enters the
+ cache, it only lives as long as self.ageLimit. This has the following implications:
+    - Accessing an object in the cache does not extend that object's lifetime in the cache
+    - When attempting to access an object in the cache that has lived longer than self.ageLimit,
+      the cache will behave as if the object does not exist
+ 
+ */
+@property (assign, getter=isTTLCache) BOOL ttlCache;
```
garrettmoon: Can you make this nonatomic and also create a getter which grabs the ivar in self.lock?

PINCache/PINMemoryCache.h
```diff
@@ -66,6 +66,16 @@ typedef void (^PINMemoryCacheObjectBlock)(PINMemoryCache *cache, NSString *key,
...
 
 /**
+ If ttlCache is YES, the cache behaves like a ttlCache. This means that once an object enters the
+ cache, it only lives as long as self.ageLimit. This has the following implications:
+ - Accessing an object in the cache does not extend that object's lifetime in the cache
+ - When attempting to access an object in the cache that has lived longer than self.ageLimit,
+ the cache will behave as if the object does not exist
+ 
+ */
+@property (assign, getter=isTTLCache) BOOL ttlCache;
```
garrettmoon: Can you make this nonatomic and write a getter which gets the ivar is self.lock?

PINCache/PINMemoryCache.m
```diff
@@ -497,10 +502,14 @@ - (void)enumerateObjectsWithBlock:(PINMemoryCacheObjectBlock)block
         return;
     
     [self lock];
+        NSDate *now = [[NSDate alloc] init];
         NSArray *keysSortedByDate = [_dates keysSortedByValueUsingSelector:@selector(compare:)];
         
         for (NSString *key in keysSortedByDate) {
-            block(self, key, _dictionary[key]);
+            // If the cache should behave like a TTL cache, then only fetch the object if there's a valid ageLimit and  the object is still alive
+            if (!self->_ttlCache || self->_ageLimit <= 0 || fabs([[_dates objectForKey:key] timeIntervalSinceDate:now]) < self->_ageLimit) {
```
garrettmoon: Since now is only used here, can you just call [NSDate date] instead?

benasher44: If I do that, we'll be reading a slightly different date (albeit very very small time delta across dates)  each time through the loop. Is that okay?

garrettmoon: Oh! Never mind, I didn't count on the loop.

benasher44: @garrettmoon well I think we're all set here then. The getters have been pushed, and I'm not sure how mark/resolve comments in github. Maybe you can't.

PINCache/PINDiskCache.m
```diff
@@ -1052,6 +1059,16 @@ - (void)setAgeLimit:(NSTimeInterval)ageLimit
     });
 }
 
+- (BOOL)isTTLCache {
```
garrettmoon: I think we need setTTLCache too? Otherwise they'll be set outside of the lock.

PINCache/PINMemoryCache.m
```diff
@@ -681,6 +690,16 @@ - (NSUInteger)totalCost
     return cost;
 }
 
+- (BOOL)isTTLCache {
```
garrettmoon: Needs a setter too?

garrettmoon:
Should ttl cache be exposed on PINCache too? Or only PINMemoryCache and PINDiskCache?

PINCache/PINDiskCache.m
benasher44: It definitely does. I'm not sure why I missed that.

benasher44:
I can't think of why you would want to use this feature on either only PINMemoryCache or PINDiskCache, but I'd say the same thing about ageLimit (also not exposed in PINCache.h), which ttlCache depends on (ttlCache = YES doesn't mean anything if ageLimit is 0). There's also some added complexity if you go that route. Let's say you do the following:
```
cache.ttlCache = YES;
cache.diskCache.ttlCache = NO;
NSLog(@"%@", cache.ttlCache ? @"YES" : @"NO");
```
What would you expect cache.ttlCache to print in this scenario? I think you avoid running into any bugs related to this situation, if you require clients to access ttlCache via either of the sub-caches.

garrettmoon:
@benasher44 you've convinced me :)

PR #68, opened 2016-02-12
Remove PIN_APP_EXTENSION requirement

benasher44:
This PR is might still be a bit rough around the edges, but everything now pod lib lints, and it also works in an extension without having to compile out the UIApplication background task bits. I could however get the ```__WATCHOS_UNAVAILABLE``` macro to work in the same way, so I just decided to screw that and compile ```TARGET_OS_WATCH``` incompatible bits out of the code.

### DIFF
--- a/PINCache/PINApplicationBackgroundTask.h
+++ b/PINCache/PINApplicationBackgroundTask.h
@@ -7,7 +7,6 @@
 //
 
 #import <Foundation/Foundation.h>
-#import <Availability.h>
 #import "PINBackgroundTask.h"
 
 #if !TARGET_OS_WATCH
--- a/PINCache/PINApplicationBackgroundTask.h
+++ b/PINCache/PINApplicationBackgroundTask.h
@@ -1,0 +1,18 @@
+//
+//  PINApplicationBackgroundTask.h
+//  PINCache
+//
+//  Created by Ben Asher on 2/11/16.
+//  Copyright © 2016 Pinterest. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import <Availability.h>
+#import "PINBackgroundTask.h"
+
+#if !TARGET_OS_WATCH
+NS_EXTENSION_UNAVAILABLE_IOS("This is a concrete PINBackgroundTask that depends on UIApplication, which cannot be used in extensions")
+@interface PINApplicationBackgroundTask : NSObject <PINBackgroundTask>
+
+@end
+#endif
--- a/PINCache/PINApplicationBackgroundTask.h
+++ b/PINCache/PINApplicationBackgroundTask.h
@@ -9,9 +9,7 @@
 #import <Foundation/Foundation.h>
 #import "PINBackgroundTask.h"
 
-#if !TARGET_OS_WATCH
 NS_EXTENSION_UNAVAILABLE_IOS("This is a concrete PINBackgroundTask that depends on UIApplication, which cannot be used in extensions")
 @interface PINApplicationBackgroundTask : NSObject <PINBackgroundTask>
 
 @end
-#endif
--- a/PINCache/PINApplicationBackgroundTask.m
+++ b/PINCache/PINApplicationBackgroundTask.m
@@ -37,10 +37,12 @@
 {
   PINApplicationBackgroundTask *task = [[self alloc] init];
 #if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+  // We have to declare this reference outside of the block, otherwise, the compiler complains about using [UIApplication sharedApplication] in a context that might be available from an extension
+  UIApplication *sharedApplication = [UIApplication sharedApplication];
   task.taskID = [[UIApplication sharedApplication] beginBackgroundTaskWithExpirationHandler:^{
     UIBackgroundTaskIdentifier taskID = task.taskID;
     task.taskID = UIBackgroundTaskInvalid;
-    [[UIApplication sharedApplication] endBackgroundTask:taskID];
+    [sharedApplication endBackgroundTask:taskID];
   }];
 #endif
   return task;
--- a/PINCache/PINApplicationBackgroundTask.m
+++ b/PINCache/PINApplicationBackgroundTask.m
@@ -6,7 +6,6 @@
 //  Copyright © 2016 Pinterest. All rights reserved.
 //
 
-#if !TARGET_OS_WATCH
 #if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
 #import <UIKit/UIKit.h>
 #endif
@@ -15,7 +14,7 @@
 
 @interface PINApplicationBackgroundTask ()
 
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0 && !TARGET_OS_WATCH
 @property (atomic, assign) UIBackgroundTaskIdentifier taskID;
 #endif
 
@@ -26,7 +25,7 @@
 - (instancetype)init
 {
   if (self = [super init]) {
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0 && !TARGET_OS_WATCH
     _taskID = UIBackgroundTaskInvalid;
 #endif
   }
@@ -36,7 +35,7 @@
 + (instancetype)start
 {
   PINApplicationBackgroundTask *task = [[self alloc] init];
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0 && !TARGET_OS_WATCH
   // We have to declare this reference outside of the block, otherwise, the compiler complains about using [UIApplication sharedApplication] in a context that might be available from an extension
   UIApplication *sharedApplication = [UIApplication sharedApplication];
   task.taskID = [[UIApplication sharedApplication] beginBackgroundTaskWithExpirationHandler:^{
@@ -50,7 +49,7 @@
 
 - (void)end
 {
-#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0 && !TARGET_OS_WATCH
   UIBackgroundTaskIdentifier taskID = self.taskID;
   self.taskID = UIBackgroundTaskInvalid;
   [[UIApplication sharedApplication] endBackgroundTask:taskID];
@@ -58,4 +57,3 @@
 }
 
 @end
-#endif
--- a/PINCache/PINApplicationBackgroundTask.m
+++ b/PINCache/PINApplicationBackgroundTask.m
@@ -1,0 +1,59 @@
+//
+//  PINApplicationBackgroundTask.m
+//  PINCache
+//
+//  Created by Ben Asher on 2/11/16.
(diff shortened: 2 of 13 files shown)


benasher44:
build_tests.sh also now pod lib lint's, so that should prevent future PRs from blocking releases.

benasher44:
Also, I couldn't find a way to do this in a way that wouldn't break existing internal behavior, so this would have to bump the version to 3.0

PINCache/PINApplicationBackgroundTask.m
```diff
@@ -0,0 +1,61 @@
...
+//  PINApplicationBackgroundTask.m
+//  PINCache
+//
+//  Created by Ben Asher on 2/11/16.
+//  Copyright © 2016 Pinterest. All rights reserved.
+//
+
+#if !TARGET_OS_WATCH
+#if __IPHONE_OS_VERSION_MIN_REQUIRED >= __IPHONE_4_0
+#import <UIKit/UIKit.h>
```
garrettmoon: Is UIKit not available on watchOS?

benasher44: Oh this looks like the 2 are related, but actually a lot of things in UIApplication.h aren't available, so I just wrapped the whole class in this.

garrettmoon:
I should have looked at comments first. Can you explain what is broken with existing internal behavior?

benasher44:
Right now, if you want to use PINCache with an extension, you compile it with a macro, and that just compiles out the parts that don't work in the extensions, and then your app and your extension can share this framework. The big downside to this is that your app can no longer use the background task behavior, since it's been compiled out. In the ideal case, you have one framework which can be used by both the extension and the app, and when your app uses it, it can still take advantage of the background task behavior. To achieve this, you have to have all methods that want to use APIs that are not available in extensions marked as NS_EXTENSION_UNAVAILABLE.

Because PINDiskCache might start creating UIApplication background tasks after init, you have to decide whether or not PINDiskCache can do this by either defining a new initializer that's marked as NS_EXTENSION_UNAVAILABLE or marking the existing initializer as NS_EXTENSION_UNAVAILABLE. When you make this decision, either one client or another is going to lose out: either previous extension clients of PINDiskCache are going to have to start calling a new extension-compatible initializer, or application clients of PINDiskCache are going to start having to call a new initializer that is marked as NS_EXTENSION_UNAVAILABLE (if they wanna take advantage of the behavior that's only available in apps). And unfortunately, this decision will also cascade up into PINCache.h as well.

benasher44:
It lints!

PINCache/PINDiskCache.m
```diff
@@ -72,44 +67,32 @@ - (instancetype)initWithName:(NSString *)name
...
+#if !TARGET_OS_WATCH
+- (instancetype)initWithBackgroundTasksWithName:(NSString *)name rootPath:(NSString *)rootPath {
+    if (!name) {
+      return nil;
+    }
+    
+    if (self = [super init]) {
+        static dispatch_once_t onceToken;
+        dispatch_once(&onceToken, ^{
+            gBackgroundTaskClass = [PINApplicationBackgroundTask class];
```
benasher44: I think I need to switch to using an ivar, since calling this initializer currently permanently alters how all PINDiskCaches behave after this call

garrettmoon:
I think I understand :) Based on what you've said it sounds like we need to have a default behavior for PINDiskCache and by proxy PINCache. Which do you think is better, background support by default or not? My feeling is it should probably use the background support APIs by default which will cause a hassle for extensions, but it will break as opposed to just not handle things correctly silently.

garrettmoon:
@benasher44 I'll try and look at this in depth tomorrow or this weekend. Are y'all blocked on this? Will getting it merged by then work?

benasher44:
@garrettmoon We're more blocked on PINCache 2.2, and hopefully #69 will get that all sorted out. This is something we definitely want soon in the coming weeks, but it's not as pressing for us. Thanks for all of the consideration :)

benasher44:
@garrettmoon okay yeah so if I go that route, if most users don't have extensions (prob the case), then the upgrade is easy. For users with extensions, they'll get a compiler error. This makes sense. I'll update later today

benasher44:
@garrettmoon In that case, do you think extensions should have an extension-safe variants of sharedCache singletons?

garrettmoon:
@benasher44 Yeah, that seems like a good call to me.

PINCache/PINDiskCache.m
```diff
@@ -243,24 +269,31 @@ +(BOOL)moveItemAtURLToTrash:(NSURL *)itemURL
...
+    NSArray *trashedItems = [[NSFileManager defaultManager] contentsOfDirectoryAtURL:[self sharedTrashURL]
+                                                          includingPropertiesForKeys:nil
+                                                                             options:0
+                                                                               error:&searchTrashedItemsError];
+    PINDiskCacheError(searchTrashedItemsError);
+    
+    for (NSURL *trashedItemURL in trashedItems) {
+      NSError *removeTrashedItemError = nil;
+      [[NSFileManager defaultManager] removeItemAtURL:trashedItemURL error:&removeTrashedItemError];
+      PINDiskCacheError(removeTrashedItemError);
```
SSheldon: Indentation here seems to have changed from 4-space to 2-space, might want to re-indent for consistency.

PINCache/PINApplicationBackgroundTask.h
```diff
@@ -0,0 +1,15 @@
+//
+//  PINApplicationBackgroundTask.h
```
SSheldon: Should this be marked as a private header? Background tasks used to be an implementation detail but now we're exposing this class and protocol publicly.

Not sure if this is a big concern, but it does seem like an implementation detail that doesn't have to be public.

PINCache/PINCache.m
```diff
@@ -72,6 +89,17 @@ + (instancetype)sharedCache
     return cache;
 }
 
++ (instancetype)sharedCacheForExtensions {
```
SSheldon: PINCache style seems to be open brackets on a newline ;)

PINCache/PINDiskCache.m
```diff
@@ -72,41 +67,24 @@ - (instancetype)initWithName:(NSString *)name
...
-        _cacheURL = [NSURL fileURLWithPathComponents:@[ rootPath, pathComponent ]];
-        
-        //we don't want to do anything without setting up the disk cache, but we also don't want to block init, it can take a while to initialize
-        //this is only safe because we use a dispatch_semaphore as a lock. If we switch to an NSLock or posix locks, this will *no longer be safe*.
-        [self lock];
-        dispatch_async(_asyncQueue, ^{
-            [self createCacheDirectory];
-            [self initializeDiskProperties];
+        _backgroundTaskClass = [PINApplicationBackgroundTask class];
+        [self prepareCacheWithName:name rootPath:rootPath];
```
SSheldon: This seems to me like a problem that should be solved with a common initializer rather than a helper method. Maybe add a private `initWithName:rootPath:backgroundTaskClass:` initializer?

Or, we could have the app-version call the extension version, like:
``` objc
if (self = [self initForExtensionsWithName:name rootPath:rootPath]) {
    _backgroundTaskClass = [PINApplicationBackgroundTask class];
}
```

PINCache/PINCache.h
```diff
@@ -101,7 +120,7 @@ typedef void (^PINCacheObjectBlock)(PINCache *cache, NSString *key, id __nullabl
  @param rootPath The path of the cache on disk.
  @result A new cache with the specified name.
  */
-- (instancetype)initWithName:(NSString *)name rootPath:(NSString *)rootPath NS_DESIGNATED_INITIALIZER;
+- (instancetype)initForExtensionsWithName:(NSString *)name rootPath:(NSString *)rootPath NS_DESIGNATED_INITIALIZER;
```
SSheldon: This will look pretty weird in Swift, `init(forExtensionsWithName:, rootPath:)`, seems like you're giving the name of the extensions that it's being initialized for. I don't have a great suggestion, I'll think on it...

PINCache/PINDiskCache.m
```diff
@@ -72,41 +67,24 @@ - (instancetype)initWithName:(NSString *)name
...
-        NSString *pathComponent = [[NSString alloc] initWithFormat:@"%@.%@", PINDiskCachePrefix, _name];
-        _cacheURL = [NSURL fileURLWithPathComponents:@[ rootPath, pathComponent ]];
-        
-        //we don't want to do anything without setting up the disk cache, but we also don't want to block init, it can take a while to initialize
-        //this is only safe because we use a dispatch_semaphore as a lock. If we switch to an NSLock or posix locks, this will *no longer be safe*.
-        [self lock];
-        dispatch_async(_asyncQueue, ^{
-            [self createCacheDirectory];
-            [self initializeDiskProperties];
+        _backgroundTaskClass = [PINApplicationBackgroundTask class];
```
SSheldon: Crazy idea: we could do something this at runtime:
```
if ([[[NSBundle mainBundle] bundlePath] hasSuffix:@".appex"]) {
    _backgroundTaskClass = nil;
} else {
    _backgroundTaskClass = NSClassFromString(@"PINApplicationBackgroundTask");
}
```

I know it's crazy and throws away some of the compile time guarantees, but jeeze Apple is it too much to ask for some way to tailor code for extensions at runtime?

PINCache/PINCache.h
benasher44: @SSheldon @garrettmoon how about ```initExtensionSafeWithName:…```?

PINCache/PINDiskCache.m
benasher44: @SSheldon I agree, but then it complains about a convenience initializer (the new initializer with the background task class arg) calling ```[super init]```, since it would then be skipping a designated initializer. To fix this, you'd have to make this new initializer designated, which isn't good since we don't want anyone to actually call it.

PINCache/PINDiskCache.m
benasher44: This seems fragile to me, since it'll break if Apple ever changes the file extension for app extensions or introduces a new app extension type with a different file extension.

benasher44: Cool idea though!

PINCache/PINCache.h
benasher44: In doing this, I could add the annotation so that this doesn't get bridged to ```init(withExtensionSafeWithName:, rootPath:)```

PINCache/PINDiskCache.m
maicki: @benasher44 
This is a quote from the ["App Extension Programming Gruide"](https://developer.apple.com/library/mac/documentation/General/Conceptual/ExtensibilityPG/ExtensionCreation.html):
> When you build an extension based on an Xcode template, you get an extension bundle that ends in .appex.

So I think checking it's fine to check for ".appex" as @SSheldon is suggesting

garrettmoon:
@benasher44 the other concern I had after seeing this (https://github.com/facebook/AsyncDisplayKit/issues/1309) is that adopting this solution will require separate APIs for projects which rely on PINCache as well.

Are you opposed to the .appex solution? @maicki put up a PR here: https://github.com/pinterest/PINCache/pull/72

I think we'd also want to at least add a extension target and unit test to make sure it doesn't break in the future.

benasher44:
@garrettmoon @maicki so my other suspicion was that this wouldn't work when PINCache is built as a framework, but I just tested that with no issues. Ship it (with tests and an extension target please)!

martin-gearzero:
For future reference, this is my preferred method for checking if the currently running process is an NSExtension which does't rely on the underlying file extension.

I think I first learned this trick [here](http://nshipster.com/new-years-2016/) under "AmIRunningAsAnExtension". Take it or leave it :)

```
+ (BOOL)isExtension {

	static BOOL isExtension;
	static dispatch_once_t onceToken;

	dispatch_once(&onceToken, ^{
		NSDictionary *extensionDictionary = [[NSBundle mainBundle] infoDictionary][@"NSExtension"];
		isExtension = [extensionDictionary isKindOfClass:[NSDictionary class]];
	});

	return isExtension;
}
```

PR #69, opened 2016-02-12
Fixed watchos linting errors

benasher44:
- travis.yaml now runs ```pod lib lint```
- Fixed up build_tests.sh to work with recent xcode tools
- Fixed some async-related test failures/flakes

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,10 @@
 language: objective-c
-script: ./build_tests.sh
+osx_image: xcode7.2
+rvm: 2.2.2
 branches:
   only:
     - master
+
+before_install: gem install cocoapods -v '0.39.0'
+before_script: pod lib lint --verbose
+script: ./build_tests.sh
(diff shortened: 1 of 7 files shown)


garrettmoon:
@benasher44 looks like pod lib lint passes, but pod spec lint fails with the same error :(

garrettmoon:
Ohhh, but it may be that it's because this isn't merged. Hold off on investigating.

benasher44:
@garrettmoon ```pod spec lint``` looks at the podspec, and downloads the code pointed to by the spec, and then it does ```pod lib lint```. Since the pod spec is still pointing at the failing tag, that's expected. Once this is merged, you'll need to update the tag. Then, it should pass.

benasher44:
This is also why I didn't have .travis.yaml do ```pod spec lint```, since it'll just be looking at the same code every time :)

garrettmoon:
@benasher44 Thanks for doing this! Let's give it a go!

garrettmoon:
Yeah, that makes sense :)

PR #72, opened 2016-03-03
Remove necessity of declaring PIN_APP_EXTENSIONS macro

maicki:


PINCache/PINDiskCache.m
```diff
@@ -1107,10 +1107,16 @@ - (void)unlock
 @end
 
 @implementation PINBackgroundTask
+
++ (BOOL)isAppExtension
+{
+    return [[[NSBundle mainBundle] executablePath] containsString:@".appex/"];
```
garrettmoon: should this be dispatch_once'd?

benasher44: Can you write a unit test for this? I think this is doable if you separated this into 2 methods, once of which just took a string and return a BOOL, and `isAppExtension` could  all that one with `[[NSBundle mainBundle] executablePath]`

benasher44:
Here's the smoke test I thought of to help ensure this can't break again:

- Add a new extension target to PINCache.xcodeproj
- Add a new test target that tests the extension target
- Add a test that just initializes the cache objects and tries to call the `start` method, which in a non-extension context would create a background task. If this breaks, the test should crash/fail.

benasher44:
Does this sound realistic/doable?

PINCache/PINDiskCache.m
maicki: Yep this will be dispatched_once. I will look into how we could integrate a unit test for it.

benasher44:
Also, @martin-gearzero's final comment in #68 seems preferable to inspecting any of the main bundle's paths.

maicki:
@benasher44 @garrettmoon Added the dispatch_once for the is app extension check and added a basic share extension target that just initializes PINCache and calls a method that would create a PINBackgroundTask.

I couldn't really figure out how we can test PINCache usage within the share extension via a test target. Happy for any help!

PINCache/PINDiskCache.m
```diff
@@ -1120,11 +1134,17 @@ - (instancetype)init
 + (instancetype)start
 {
     PINBackgroundTask *task = [[self alloc] init];
```
maicki: @garrettmoon You think we should just return nil here and not allocate any PINBackgroundTask if isAppExtension == YES?

garrettmoon: Seems like a good call to me.

garrettmoon:
@maicki It might be possible to do with a UI automation test? Force the extension to run in Safari via a share extension?

I think this is a good start though and worth merging.

PR #74, opened 2016-03-15
Removed local weak self references; It is the same as having one weak…

Amosel:
Removed unnecessary local weak self references.


garrettmoon:
Indeed! Thanks!

PR #76, opened 2016-03-23
Nullability correction

ejensen:
PINDiskCache's `fileURLForKey:` returns nil whenever the cache does not hold the object for the specified key.


garrettmoon:
Thanks!

Issue #81, opened 2016-04-21
PINCache does not support file protection


sebastienvillar:
All files are currently written unencrypted on disk.
For protection of the user's privacy, files should be encrypted when needed if necessary

PR #82, opened 2016-04-21
Add support for file writing protection

sebastienvillar:
This PR adds file protection to files stored in the PINDiskCache.
The related issue is https://github.com/pinterest/PINCache/issues/81

### DIFF
--- a/PINCache/PINDiskCache.h
+++ b/PINCache/PINDiskCache.h
@@ -99,6 +99,21 @@ typedef void (^PINDiskCacheObjectBlock)(PINDiskCache *cache, NSString *key, id <
  */
 @property (assign) NSTimeInterval ageLimit;
 
+
+/**
+ The writing protection option used when writing a file on disk. This value is used every time an object is set.
+ NSDataWritingAtomic and NSDataWritingWithoutOverwriting are ignored if set
+ Defaults to NSDataWritingFileProtectionNone.
+ 
+ @warning Only new files are affected by the new writing protection. If you need all files to be affected,
+ you'll have to purge and set the objects back to the cache
+ 
+ Only available on iOS
+ */
+#if TARGET_OS_IPHONE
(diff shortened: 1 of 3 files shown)


PINCache/PINDiskCache.h
```diff
@@ -1,5 +1,6 @@
 //  PINCache is a modified version of TMCache
 //  Modifications by Garrett Moon
+//  Modifications by Sebastien Villar, Microsoft
```
garrettmoon: Do you mind taking this out? If it feels better to you, you can take out the Modifications by Garrett Moon too. I put that there when I forked the project originally because I modified nearly 100% of the project (that was my justification). But I'd rather not see this turn into a list of contributors. It would get, I hope, very long :)

You can take solace that you're name will show up in contributors and in the git blame :P

tests/PINCacheTests/PINCacheTests.m
```diff
@@ -1,5 +1,6 @@
 //  PINCache is a modified version of TMCache
 //  Modifications by Garrett Moon
+//  Modifications by Sebastien Villar, Microsoft
```
garrettmoon: Same here as above.

garrettmoon:
Cool improvement!

garrettmoon:
Looks like the build is failing because there's no support for NSDataWritingFileProtectionNone on OS X?

PINCache/PINDiskCache.h
sebastienvillar: Sure I can definitely remove it. I thought I needed that for compliance but it's fine :S

sebastienvillar:
Thanks!
You're right. Data protection is enabled at the system level with FileVault.
I've removed this field for OS X

garrettmoon:
Awesome, thanks so much for adding this! Cool feature!

PR #84, opened 2016-05-01
switch PINCacheObjectSubscripting.h from project to public

heyalexchoi:
exposes PINCacheObjectSubscripting.h in framework and fixes this build error:

screen shot 2016-05-01 at 5 45 54 pm


garrettmoon:
I'm sorry, I didn't get this merged earlier, dropped the ball. Can you confirm this isn't the case now in master? I believe they are now public. Lemme know if they're not and I'll re-open.

Issue #79, opened 2016-04-09
checking if image is cached


DanielRanLehmann:
I might be missing something obvious. I'm using PINcache to store my images that is async plugged into a custom cell in a collectionView.

How can I check if an image is cached?

Is there any similar method to SDWebImage where you can check if an image is cached on disk like this:
- (BOOL)diskImageExistsForURL:(NSURL *)url;

- Daniel Lehmann

garrettmoon:
@DanielRanLehmann You can obviously just call objectForKey, but that will instantiate the object off disk which if you just want to check if it's available is expensive. Instead, you can access the memoryCache and diskCache directly. objectForKey on the memory cache will determine existence (and isn't expensive). fileURLForKey is the cheapest way to check existence on the disk cache.

I'd be happy to accept a pull request which exposes methods on the memoryCache, diskCache and PINCache to essentially do this all.

DanielRanLehmann:
Thanks for the quick response :+1: I will look into it right away and make a pull request if it works out..

danielgalasko:
@garrettmoon I think this can be closed:)

garrettmoon:
Thanks!

PR #85, opened 2016-05-08
Add API to determines whether an object is present for the given key in the cache

maicki:
Adresses Issue #79

### DIFF
--- a/PINCache/PINCache.m
+++ b/PINCache/PINCache.m
@@ -74,6 +74,22 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
 
 #pragma mark - Public Asynchronous Methods -
 
+- (void)containsObjectForKey:(NSString *)key block:(PINCacheObjectContainmentBlock)block
+{
+    if (!key || !block) {
+        return;
+    }
+    
+    __weak PINCache *weakSelf = self;
+    
+    dispatch_async(_concurrentQueue, ^{
+        PINCache *strongSelf = weakSelf;
+        
+        BOOL containsObject = [strongSelf containsObjectForKey:key];
+        block(containsObject);
+    });
+}
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wshadow"
 
@@ -292,6 +308,14 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     }];
     
     return byteCount;
+}
+
+- (BOOL)containsObjectForKey:(NSString *)key
+{
+    if (!key)
+        return NO;
+    
+    return [_memoryCache containsObjectForKey:key] || [_diskCache containsObjectForKey:key];
 }
 
 - (__nullable id)objectForKey:(NSString *)key
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -459,17 +459,14 @@ static NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
 
 - (void)containsObjectForKey:(NSString *)key block:(PINDiskCacheContainsBlock)block
 {
+    if (!key || !block)
+        return;
+    
     __weak PINDiskCache *weakSelf = self;
     
     dispatch_async(_asyncQueue, ^{
         PINDiskCache *strongSelf = weakSelf;
-        BOOL containsObject = [strongSelf containsObjectForKey:key];
-        
-        if (block) {
-            [strongSelf lock];
-                block(containsObject);
-            [strongSelf unlock];
-        }
+        block([strongSelf containsObjectForKey:key]);
     });
 }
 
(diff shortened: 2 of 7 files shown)


PINCache/PINCache.m
```diff
@@ -294,6 +310,14 @@ - (NSUInteger)diskByteCount
     return byteCount;
 }
 
+- (BOOL)containsObjectForKey:(NSString *)key
+{
+    if (!key)
+        return NO;
```
jparise: Braces for consistency?

PINCache/PINCache.m
```diff
@@ -74,6 +74,22 @@ + (instancetype)sharedCache
 
 #pragma mark - Public Asynchronous Methods -
 
+- (void)containsObjectForKey:(NSString *)key block:(PINCacheObjectContainmentBlock)block
+{
+    if (!key || !block) {
```
jparise: The `!block` early-return condition makes sense, but I'm not sure about `!key` not executing the block. It's such an exceptional case that I don't think it's that important in practice, but I also think it's a stronger contract to say that the block is always invoked if one is provided.

PINCache/PINDiskCache.m
```diff
@@ -457,6 +457,22 @@ - (void)lockFileAccessWhileExecutingBlock:(void(^)(PINDiskCache *diskCache))bloc
     });
 }
 
+- (void)containsObjectForKey:(NSString *)key block:(PINDiskCacheContainsBlock)block
+{
+    __weak PINDiskCache *weakSelf = self;
+    
+    dispatch_async(_asyncQueue, ^{
+        PINDiskCache *strongSelf = weakSelf;
+        BOOL containsObject = [strongSelf containsObjectForKey:key];
```
jparise: Move this call within the `if (block)` check? Otherwise, this is wasted work.

PINCache/PINDiskCache.m
```diff
@@ -457,6 +457,22 @@ - (void)lockFileAccessWhileExecutingBlock:(void(^)(PINDiskCache *diskCache))bloc
...
+        
+        if (block) {
+            [strongSelf lock];
```
jparise: Maybe I'm overlooking something, but why do we need to lock on reporting the result? Should the lock instead of around the `-containsObjectForKey:` call?

PINCache/PINMemoryCache.m
```diff
@@ -270,6 +270,22 @@ - (void)trimToAgeLimitRecursively
 
 #pragma mark - Public Asynchronous Methods -
 
+- (void)containsObjectForKey:(NSString *)key block:(PINMemoryCacheContainmentBlock)block
+{
+    if (!key || !block) {
```
jparise: Similar comment as for the disk cache.

PINCache/PINCache.m
maicki: We don't use the braces in other cases or do I overlook something?

jparise: You do above on Line 79-81.

maicki: Will change in line 79-81 and other places that I added braces to be consistent across the project.

PINCache/PINDiskCache.m
maicki: We actually don't need a lock in here as we already lock in `containsObjectForKey:`. Will change.

PINCache/PINCache.m
maicki: If you look into the header file you can see that we declare key and block as `nonnull`. So currently the contract explicitly requires a key and block and the API consumer should not pass nil in one of the two arguments. We could discuss changing the block to `nullable`, but as the method does not have a side effect like calling `setObject:forKey:block:` it makes no really sense to call the method without a key or block argument. I decided to define both arguments `nonnull`. In this case we can early return if no key or block is given to prevent bad behavior if called from within Obj-C, Swift should already prevent API consumer to call it with a nil parameter in the first place.

maicki:
I addressed @jparise comments. Can you take another look please. Thanks!

cc @jparise @garrettmoon

PINCache/PINCache.m
jparise: Thanks, that makes sense. I missed those header annotations.

jparise:
This looks good to me. Test coverage could be expanded a bit more to cover all of the edge cases, but that's probably true in general.

@garrettmoon for any final comments.

garrettmoon:
Going to take a final look at this today, thanks @maicki and @jparise !

tests/PINCacheTests/PINCacheTests.m
```diff
@@ -136,6 +136,27 @@ - (void)testObjectSetWithDuplicateKey
...
 
+- (void)testObjectContains
+{
+    NSString *key = @"key";
+    NSString *value = @"value";
+    
+    [self.cache setObject:value forKey:key];
+    
+    // Synchronously
+    XCTAssertTrue([self.cache containsObjectForKey:@"key"], @"object was gone");
```
garrettmoon: nit: use key local var instead of `@"key"`

garrettmoon:
Tests passed locally, so I'm merging. Dunno what's going on with Travis :/

PR #86, opened 2016-05-08
Fix Xcode warnings printing a pointer value for self

maicki:


garrettmoon:
Tests passed locally, merging.

PR #90, opened 2016-05-30
added tvOS framework target

ittna:
Added tvOS target to project file. Carthage needs it for building tvOS version of the library.


garrettmoon:
I totally forgot about this, sorry :( Merging now!

PR #96, opened 2016-07-19
Use pthread mutex instead of a semaphore for locking

maicki:
For performance reasons move locking from a `dispatch_semaphore_t` to `pthread_mutex_t`.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -8,6 +8,8 @@
 #import <UIKit/UIKit.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


garrettmoon:
I *think* that something like this would be a bit more performant? https://github.com/pinterest/PINRemoteImage/blob/master/Pod/Classes/PINRemoteLock.m

I'd also be ok with using a mutex directly.

jparise:
Can you update the commit description to explain the benefit of this change?

PR #97, opened 2016-07-21
Remove locking on callback

garrettmoon:
Previously, we would always hold the PINDiskCache lock when calling callbacks. This can result in two classes of errors.
- deadlocks by calling synchronous methods within the callbacks.
- stalling of the disk cache when long running work is done within the callbacks.

This change makes it so the standard callback block no longer returns the fileURL and therefore no longer needs to continue holding the lock. It also documents the fileURL methods to indicate that they should be used with care.


PINCache/PINDiskCache.m
```diff
@@ -113,8 +113,8 @@ - (instancetype)initWithName:(NSString *)name rootPath:(NSString *)rootPath
         //this is only safe because we use a dispatch_semaphore as a lock. If we switch to an NSLock or posix locks, this will *no longer be safe*.
```
rahul-malik: Didn't we switch to an NSLock in a recent pr from @maicki ? Might make sense to update this comment

garrettmoon: Address in #98 Thanks!

PR #99, opened 2016-07-22
Remove background tasks

garrettmoon:
@johnezang and @appleguy brought up issues with background tasks and
performance (specifically in our app we were seeing the creation of the
tasks take a significant portion of all CPU time on single core devices).

Background tasks are created for the following actions:

- Emptying the trash
- Setting an object
- Removing an object
- Trimming the cache (size, date, size/date)
- Removing all objects
- Enumerating all objects

In the above list, only one task is truly necessary to complete to
avoid getting in an inconsistent state in a subsequent run, setObject:

After reasoning about it for  some time, I believe that creating
the background tasks is no longer necessary because of our try catch
on object unarchive.

The background tasks mainly ensure the cache stays in a consistent state
however they cannot guarantee this as an app can crash or be killed at
any time.

This also has the added advantage of simplifying app extensions which
never supported this in the first place.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -17,14 +17,6 @@ __LINE__, [error localizedDescription]); }
 static NSString * const PINDiskCachePrefix = @"com.pinterest.PINDiskCache";
 static NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
 
-@interface PINBackgroundTask : NSObject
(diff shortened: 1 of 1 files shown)


jparise:
Is there a way we can "prove" the correctness of the cache after this change, perhaps through some sort of automated stress test?

garrettmoon:
@jparise I'm not sure how we would, do you have ideas? Only way to simulate what's happening is to essentially kill the disk cache in progress.

garrettmoon:
Going to land this for now, would like to get it in the next release. I'm happy to work with someone figuring out an automated stress test, seems like it could be invaluable.

PR #101, opened 2016-08-05
Add fileURL back into `PINDiskCacheObjectBlock` and allow custom file extension

alexanderedge:
Hi there,

Firstly, I'd like to say thank you for the work you've put into maintaining PINCache. I've started using this instead of TMDiskCache for storing video clips.

So, this PR consists of two parts:

## Adding the fileURL back into `PINDiskCacheObjectBlock`

I'm not sure what the rationale behind its removal was, but in my use-case it would require an additional call to the cache after adding an object in order to retrieve the fileURL. It seems unnecessary since the fileURL is available at the time of storage.

## Adding an optional file extension

By default, files are stored using an escaped key. If I provide a serialiser / deserialiser that doesn't perform any operation on `NSData` objects, I can feed the file URLs directly into other API, in my case `AVPlayerItem`. Unfortunately, `AVPlayerItem` is particularly strict when it comes to reading file URLs and requires that there be an extension.

I have therefore added a `fileExtension` property to `PINDiskCache` that is used after encoding a key.

I appreciate that you may not wish to merge these commits as they are specific to my use of `NSData` and `AVPlayerItem`, however, I thought I'd give you the opportunity to review them in case they are of use.

### DIFF
--- a/PINCache/PINCache.m
+++ b/PINCache/PINCache.m
@@ -122,7 +122,7 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
                         block(strongSelf, memoryCacheKey, memoryCacheObject);
                 });
             } else {
-                [strongSelf->_diskCache objectForKey:memoryCacheKey block:^(PINDiskCache *diskCache, NSString *diskCacheKey, id <NSCoding> diskCacheObject) {
+                [strongSelf->_diskCache objectForKey:memoryCacheKey block:^(PINDiskCache *diskCache, NSString *diskCacheKey, id <NSCoding> diskCacheObject, NSURL *fileURL) {
                     PINCache *strongSelf = weakSelf;
                     if (!strongSelf)
                         return;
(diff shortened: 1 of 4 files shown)


garrettmoon:
Hi @alexanderedge thanks for putting up a PR!

With regards to why I removed the fileURL from the callback: Essentially, for the fileURL to be valid, PINDiskCache's lock must be held, this meant that every callback from the diskCache would block any other work being done with the disk cache. It also meant that it was easier for a developer to cause a deadlock by calling a synchronous method on the disk cache (or by proxy a cache that owned the disk cache).

So that was the reasoning (and I believe is documented in the headers and the commit message of the change).

In the end, I believe retrieving the fileURL from the diskCache is actually relatively cheap. We could possibly invest in making a method which took a UUID or some other piece of data that the callback could return to make it even cheaper if that's your concern?

If possible, I'd really appreciate separating out the custom file extension stuff into a separate PR so each change can be discussed and reviewed on its own.

alexanderedge:
Thanks for the clarification. I'm happy to keep the block signature as it was; I don't think it's worth investigating a workaround for a potential deadlock if we can avoid it altogether.

I have opened #102 for the file extension.

PR #102, opened 2016-08-05
Add support for a custom file extension

alexanderedge:
Separated from #101 

By default, files are stored using an escaped key. If I provide a serialiser / deserialiser that doesn't perform any operation on NSData objects, I can feed the file URLs directly into other API, in my case AVPlayerItem. Unfortunately, AVPlayerItem is particularly strict when it comes to reading file URLs and requires that there be an extension.

I have added a fileExtension property to PINDiskCache that is used after encoding a key.

### DIFF
--- a/PINCache/PINDiskCache.h
+++ b/PINCache/PINDiskCache.h
@@ -129,7 +129,7 @@ typedef id<NSCoding> __nonnull(^PINDiskCacheDeserializerBlock)(NSData* data);
 /**
  Extension for all cache files on disk. Defaults to no extension. 
  */
-@property (nonatomic, nullable, copy) NSString *fileExtension;
+@property (readonly) NSString *fileExtension;
 
 /**
  The writing protection option used when writing a file on disk. This value is used every time an object is set.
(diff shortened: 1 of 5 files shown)


PINCache/PINDiskCache.m
```diff
@@ -187,7 +187,13 @@ - (NSString *)encodedString:(NSString *)string
     }
     
     if ([string respondsToSelector:@selector(stringByAddingPercentEncodingWithAllowedCharacters:)]) {
-        return [string stringByAddingPercentEncodingWithAllowedCharacters:[[NSCharacterSet characterSetWithCharactersInString:@".:/%"] invertedSet]];
+        NSString *encodedString = [string stringByAddingPercentEncodingWithAllowedCharacters:[[NSCharacterSet characterSetWithCharactersInString:@".:/%"] invertedSet]];
+        if (self.fileExtension) {
```
jparise: `self.fileExtension.length > 0` ?

PINCache/PINDiskCache.h
```diff
@@ -126,6 +126,10 @@ typedef id<NSCoding> __nonnull(^PINDiskCacheDeserializerBlock)(NSData* data);
  */
 @property (assign) NSTimeInterval ageLimit;
 
+/**
+ Extension for all cache files on disk. Defaults to no extension. 
+ */
+@property (nonatomic, nullable, copy) NSString *fileExtension;
```
garrettmoon: Should this need to be set on init? Is there any reason you would want to change this after init?

If not, this needs a getter and setter implemented which hold onto the lock. See `willAddObjectBlock` and `setWillAddObjectBlock:`

garrettmoon:
@alexanderedge can you get that last self.fileExtension -> self.filextensionLength that Jon pointed out. Will merge after that :)

alexanderedge:
Apologies, missed that one. Thanks for taking the time to review, I appreciate it.

garrettmoon:
No worries, thank you for taking the time to put up the PR and improve the project!

PR #104, opened 2016-08-09
Avoid holding lock while serializing / deserializing

garrettmoon:
No need to hold the lock, and depending on what the client
is doing, they could be serializing / deserializing for a long
time.


chrisdanford:
LGTM, FWIW.

garrettmoon:
Thanks @chrisdanford :)

PR #106, opened 2016-08-25
[PINMemoryCache] Replace Semaphore with pthread_mutex

Adlai-Holler:
@garrettmoon before you ask… yes, a semaphore killed my parents. 🙃

### DIFF
--- a/PINCache/PINCache.h
+++ b/PINCache/PINCache.h
@@ -209,7 +209,7 @@ typedef void (^PINCacheObjectContainmentBlock)(BOOL containsObject);
 
 /**
  Retrieves the object for the specified key. This method blocks the calling thread until the object is available.
- Uses a semaphore to achieve synchronicity on the disk cache.
+ Uses a lock to achieve synchronicity on the disk cache.
  
  @see objectForKey:block:
  @param key The key associated with the object.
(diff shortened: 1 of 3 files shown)


maicki:
@Adlai-Holler See #98 maybe we should use the same in this case.

garrettmoon:
The conditional lock is required in the disk case because of the weird initialization we do :/ I think we can use a semaphore directly here. Eventually I'd like to use something that uses an NSLock in debug and a semaphore in production ala the lock in PINRemoteImage.

Adlai-Holler:
I think we should prefer a mutex over a semaphore here. Mutex was built exactly for this case – for one thread to own and block all other threads from acquiring. It's the basis of NSLock. I don't know the implementation details between mutex and semaphore, maybe @johnezang can weight in.

garrettmoon:
@Adlai-Holler Sorry, I meant NSLock in debug, mutex in production :(

garrettmoon:
I'm happy to merge this once the tests pass.

Adlai-Holler:
Oh! Awesome!

garrettmoon:
Thanks @Adlai-Holler !

levi:
semaphores caused Y2K

PR #108, opened 2016-08-28
Adds PINOperationQueue

garrettmoon:
PINOperationQueue is an operation queue designed to allow
specifying the maximum number of threads, support priority (ordering,
not thread priority and cancelation).

It is also designed to prevent live locks due to priority inversion.
It attempts to achieve this by maintaining two queues: a serial queue
which always takes operations in FIFO order and a concurrent queue
which runs operations in priority order.

### DIFF
--- a/PINCache/PINOperationQueue.m
+++ b/PINCache/PINOperationQueue.m
@@ -1,0 +1,217 @@
+//
+//  PINOperationQueue.m
+//  Pods
+//
+//  Created by Garrett Moon on 8/23/16.
+//
+//
+
+#import "PINOperationQueue.h"
+#import <pthread.h>
+
+@interface NSNumber (PINOperationQueue) <PINOperationReference>
+
+@end
+
+@interface PINOperationQueue () {
+  pthread_mutex_t _lock;
+  //increments with every operation to allow cancelation
+  NSUInteger _operationReferenceCount;
+  
+  dispatch_queue_t _serialQueue;
+  BOOL _serialQueueBusy;
+  
+  dispatch_semaphore_t _concurrentSemaphore;
+  dispatch_queue_t _concurrentQueue;
+  dispatch_queue_t _semaphoreQueue;
+  
+  NSMutableOrderedSet *_allOperations;
+  NSMutableOrderedSet *_lowPriorityOperations;
+  NSMutableOrderedSet *_defaultPriorityOperations;
+  NSMutableOrderedSet *_highPriorityOperations;
+  
+  NSHashTable *_canceledOperations;
+}
+
+@end
+
+@interface PINOperation : NSObject
+
+@property (nonatomic, strong) dispatch_block_t block;
+@property (nonatomic, strong) id <PINOperationReference> reference;
+
++ (instancetype)operationWithBlock:(dispatch_block_t)block reference:(id <PINOperationReference>)reference;
+
+@end
+
+@implementation PINOperation
+
++ (instancetype)operationWithBlock:(dispatch_block_t)block reference:(id<PINOperationReference>)reference
+{
+  PINOperation *operation = [[self alloc] init];
+  operation.block = block;
+  operation.reference = reference;
+
+  return operation;
+}
+
+@end
+
+@implementation PINOperationQueue
+
+- (instancetype)initWithMaxConcurrentOperations:(NSUInteger)maxConcurrentOperations
+{
+  if (self = [super init]) {
+    NSAssert(maxConcurrentOperations > 1, @"Max concurrent operations must be greater than 1. If it's one, just use a serial queue!");
+    _operationReferenceCount = 0;
+    
+    pthread_mutexattr_t attr;
+    pthread_mutexattr_init(&attr);
+    pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
+    pthread_mutex_init(&_lock, &attr);
+    
+    _serialQueue = dispatch_queue_create("PINOperationQueue Serial Queue", DISPATCH_QUEUE_SERIAL);
+    
+    _concurrentQueue = dispatch_queue_create("PINOperationQueue Unprioritized Serial Queue", DISPATCH_QUEUE_CONCURRENT);
+    _concurrentSemaphore = dispatch_semaphore_create(maxConcurrentOperations - 1);
+    _semaphoreQueue = dispatch_queue_create("PINOperationQueue Serial Semaphore Queue", DISPATCH_QUEUE_SERIAL);
+    
+    _allOperations = [[NSMutableOrderedSet alloc] init];
+    _lowPriorityOperations = [[NSMutableOrderedSet alloc] init];
+    _defaultPriorityOperations = [[NSMutableOrderedSet alloc] init];
+    _highPriorityOperations = [[NSMutableOrderedSet alloc] init];
+  }
+  return self;
+}
+
+- (void)dealloc
+{
+  pthread_mutex_destroy(&_lock);
+}
+
+- (id <PINOperationReference>)nextOperationReference
+{
+  [self lock];
+    id <PINOperationReference> reference = [NSNumber numberWithUnsignedInteger:++_operationReferenceCount];
+  [self unlock];
+  return reference;
+}
+
+- (id <PINOperationReference>)addOperation:(dispatch_block_t)block withPriority:(PINOperationQueuePriority)priority
+{
+  id <PINOperationReference> reference = [self nextOperationReference];
+  
+  NSMutableOrderedSet *queue;
+  switch (priority) {
+    case PINOperationQueuePriorityLow:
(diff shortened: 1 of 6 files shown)


PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+    id <PINOperationReference> reference = [NSNumber numberWithUnsignedInteger:++_operationReferenceCount];
+  [self unlock];
+  return reference;
+}
+
+- (id <PINOperationReference>)addOperation:(dispatch_block_t)block withPriority:(PINOperationQueuePriority)priority
+{
+  id <PINOperationReference> reference = [self nextOperationReference];
+  
+  NSMutableOrderedSet *queue;
```
johnepinterest: This should be initialized to `NULL` to prevent accidental use of an uninitialized value.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+  switch (priority) {
```
johnepinterest: This `switch` statement should have a `default:` case.  If, for whatever reason, `priority` has a value other than the enumerated cases, this would cause `queue` to not be set, allowing the use of an uninitialized value.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+    [queue addObject:operation];
+    [_allOperations addObject:operation];
+  [self unlock];
+  
+  [self scheduleOperations:NO];
+  
+  return reference;
+}
+
+- (void)cancelOperation:(id <PINOperationReference>)operationReference
```
johnepinterest: This doesn't seem to do anything?  I don't see how the operation is ever canceled.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+{
+  [self lock];
+    [_canceledOperations addObject:operationReference];
```
johnepinterest: There doesn't seem to be anything that processes `_canceledOperations`, so it's just going to grow and grow.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+          dispatch_semaphore_signal(_concurrentSemaphore);
+        });
+      } else {
+        dispatch_semaphore_signal(_concurrentSemaphore);
+      }
+  });
+}
+
+//Call with lock held
+- (PINOperation *)nextOperationByPriority
```
johnepinterest: Should be modified to always acquire the objects lock.  It's safer and doesn't violate the principle of least surprise for the next person working on the code.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+  operation = [_defaultPriorityOperations firstObject];
+  if (operation) {
+    return operation;
+  }
+  operation = [_lowPriorityOperations firstObject];
+  return operation;
+}
+
+//Call with lock held
+- (void)removeOperation:(PINOperation *)operation
```
johnepinterest: Should be modified to always acquire the objects lock.  It's safer and doesn't violate the principle of least surprise for the next person working on the code.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+}
+
+- (void)cancelOperation:(id <PINOperationReference>)operationReference
+{
+  [self lock];
+    [_canceledOperations addObject:operationReference];
+  [self unlock];
+}
+
+- (void)scheduleOperations:(BOOL)onlyCheckSerial
```
johnepinterest: It seems like the way that it's written, this will actually execute `maxConcurrentOperations + 1`? The reason I say this is it seems like it will run `maxConcurrentOperations` on the `concurrentQueue` plus one on the `serialQueue`?  Or did I miss something?

johnepinterest: I found it extremely hard to reason about exactly what happens in here.  To be clear, I believe the `serialQueue` / `concurrentQueue` priority inversion avoidance logic is sound and works, but I found it difficult to convince myself that if there's an operation that's enqueued that it is guaranteed that it will be executed.  I didn't find a case where it could happen, but I also can't say that it's impossible either.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+@interface PINOperationQueue () {
+  pthread_mutex_t _lock;
+  //increments with every operation to allow cancelation
+  NSUInteger _operationReferenceCount;
+  
+  dispatch_queue_t _serialQueue;
+  BOOL _serialQueueBusy;
+  
+  dispatch_semaphore_t _concurrentSemaphore;
+  dispatch_queue_t _concurrentQueue;
```
appleguy: Do we need to use a separate queue instead of the system global concurrent queue?

Is there any way to avoid a third queue for the semaphore?  Haven't fully reviewed the code yet (my bad), just thinking that this could be a really fantastic primitive to make wider use of in cases we may create a lot of them (like layout systems) and curious if any of these can be safely unified.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,217 @@
...
+  [self unlock];
+  
+  if (onlyCheckSerial) {
+    return;
+  }
+  
+  dispatch_async(_semaphoreQueue, ^{
+      dispatch_semaphore_wait(_concurrentSemaphore, DISPATCH_TIME_FOREVER);
+      [self lock];
+        PINOperation *operation = [self nextOperationByPriority];
```
aurelienp: I'd expect a method called `scheduleOperations:` to schedule all operations. Unless I'm wrong, you schedule only the top priority one here.
It might bite you later if you decide to add a `addOperations:withPriority:` method because you'd reasonably call `scheduleOperations:` once.

aurelienp:
@garrettmoon, nice work on this PR! 

I'm curious though, could you tell me if you faced issues with `NSOperationQueue`, `NSOperationQueue.maxConcurrentOperationCount` and `NSOperation.queuePriority` that made you switch to this home-made approach? Potential live-locks? Opaque priority management? Others?

appleguy:
We measured a lot of overhead from NSOperation, its use of KVO (even if no observers outside the framework) being one of the causes.  The lack of knowledge of its prioritization techniques also made it difficult to trust with a user-experience-critical system like image downloading, because on a slow connection it can make a big difference if the most needed image gets queued because the last slot was assigned without careful consideration :). Ironically the two relate as changing the priority triggers a lot of KVO cost, and because ASDK provides a ton of signals about what is going on at the UI layer, we want to be able to rapidly adjust priority for negligible cost.

PINCache/PINOperationQueue.m
aurelienp: > this will actually execute maxConcurrentOperations + 1?

I think @garrettmoon covers this when he creates the semaphore:
`_concurrentSemaphore = dispatch_semaphore_create(maxConcurrentOperations - 1);`

aurelienp:
This is very clear and interesting feedback @appleguy, thank you.

PINCache/PINOperationQueue.m
garrettmoon: @aurelienp exactly, I'll add a comment there.

garrettmoon: @johnepinterest That's honestly where I am as well with the priority inversion. I'd love to hear if you have a better idea on how to implement this reliably.

PINCache/PINOperationQueue.m
garrettmoon: @appleguy Honestly, I'm not 100% sure of all the tradeoffs around using the global queue vs a new one. One issue I could see with using the global queue is starvation. I guess I could make it an init option.

PINCache/PINOperationQueue.m
garrettmoon: I updated to rename them with the prefix `locked_` hopefully anyone coming along after wouldn't be surprised that way :)

PINCache/PINOperationQueue.m
garrettmoon: I'd forgotten about this D: Implemented (with a test) now.

PINCache/PINOperationQueue.m
```diff
@@ -0,0 +1,246 @@
...
+{
+  id <PINOperationReference> reference = [self nextOperationReference];
+  
+  NSMutableOrderedSet *queue = nil;
+  switch (priority) {
+    case PINOperationQueuePriorityLow:
+      queue = _lowPriorityOperations;
+      break;
+      
+    default:
```
aurelienp: ^

garrettmoon: ?

garrettmoon: Aha :) I have another PR that's going up right after this one lands to refactor this out into a separate method. I'll fix it there.

PR #111, opened 2016-09-04
Expose key for custom serializer and deserializer

sainttail:
Expose the key on custom serializer. This way we can do any custom serialization based on different key.


garrettmoon:
Thanks! Looks great!

PR #116, opened 2016-10-03
Significantly improves PINDiskCache performance. ~20% improvement in …

garrettmoon:
…Pinterest.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -168,12 +168,14 @@ typedef NS_ENUM(NSUInteger, PINDiskCacheCondition) {
 
 #pragma mark - Private Methods -
 
-- (NSURL *)_locked_encodedFileURLForKey:(NSString *)key
+- (NSURL *)encodedFileURLForKey:(NSString *)key
 {
     if (![key length])
         return nil;
     
-    return [_cacheURL URLByAppendingPathComponent:[self encodedString:key]];
+    //Significantly improve performance by indicating that the URL will *not* result in a directory.
+    //Also note that accessing _cacheURL is safe without the lock because it is only set on init.
+    return [_cacheURL URLByAppendingPathComponent:[self encodedString:key] isDirectory:NO];
 }
 
 - (NSString *)keyForEncodedFileURL:(NSURL *)url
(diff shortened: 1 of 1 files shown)


PINCache/PINDiskCache.m
```diff
@@ -771,14 +778,21 @@ - (NSURL *)fileURLForKey:(NSString *)key updateFileModificationDate:(BOOL)update
...
     
     [self lock];
-        fileURL = [self _locked_encodedFileURLForKey:key];
-        
         if ([[NSFileManager defaultManager] fileExistsAtPath:[fileURL path]]) {
             if (updateFileModificationDate) {
-                [self _locked_setFileModificationDate:now forURL:fileURL];
+                //It's unlikely these need to be run in order or to be run quickly. Get it done eventually instead of blocking.
+                __weak PINDiskCache *weakSelf = self;
+                dispatch_async(_asyncQueue, ^{
```
appleguy: Seems worth factoring out this sub-block into a method, maybe including the dispatch_async?

appleguy:
@garrettmoon the build is failing, but I know why.  xcpretty was updated to 0.2.3 and it also broke AsyncDisplayKit.  Reverting to 0.2.2 fixes the problem.

Also, tests are running on iOS 8.4 on iPhone 6 — maybe we should update these to iOS 10 / iPhone 7 now?

screen shot 2016-10-03 at 8 21 57 pm

appleguy:
@garrettmoon here's the xcpretty change https://github.com/facebook/AsyncDisplayKit/commit/7fc9f1824ac327cb5a3a7a1b7b77e6d0920fc3ff

PINCache/PINDiskCache.m
```diff
@@ -720,15 +722,11 @@ - (id)objectForKeyedSubscript:(NSString *)key
...
     id <NSCoding> object = nil;
-    NSURL *fileURL = nil;
+    NSURL *fileURL = [self encodedFileURLForKey:key];
     
     [self lock];
-        fileURL = [self _locked_encodedFileURLForKey:key];
-        object = nil;
-        
-        if ([[NSFileManager defaultManager] fileExistsAtPath:[fileURL path]] &&
+        if (!self->_ttlCache || self->_ageLimit <= 0 || fabs([[_dates objectForKey:key] timeIntervalSinceDate:now]) < self->_ageLimit) {
```
chrisdanford: Is removing the `fileExistsAtPath` going be an OK behavior change?

PINCache/PINDiskCache.m
```diff
@@ -744,9 +742,18 @@ - (id)objectForKeyedSubscript:(NSString *)key
                 PINDiskCacheError(error);
             }
             [self lock];
-          if (!self->_ttlCache) {
-            [self _locked_setFileModificationDate:now forURL:fileURL];
```
chrisdanford: - There's one call to `_locked_setFileModificationDate` in `setObject:forKey:fileURL:` that you didn't make async.  Intentional?
- If there are two copies of the same ~10 lines to set this async, worth breaking into a separate method?

PINCache/PINDiskCache.m
garrettmoon: Yes because initWithContentsOfFile returns nil if the file doesn't exist / doesn't throw an exception.

PINCache/PINDiskCache.m
garrettmoon: I thought I had a good reason, but thinking through it more I don't. I'll pull it out into another method and call it from every place, good feedback.

garrettmoon: Ok, actually perf appears to be worse in the set case because of the extra dispatchs clogging everything up, so I abstracted it out in those two cases.

PINCache/PINDiskCache.m
```diff
@@ -847,6 +842,7 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key fileURL:(NSURL **
         PINDiskCacheError(writeError);
         
         if (written) {
+            //Attempting to do this asynchronously seems to result in *worse* performance.
             [self _locked_setFileModificationDate:now forURL:fileURL];
```
garrettmoon: I actually wondered about that too and then forgot to check. I'll do that.

garrettmoon:
Actual measured gains on an iPhone 4:

Percentage decrease:
Set: 8%
Get Hit: 60%
Get Miss: 38%

PR #118, opened 2016-10-06
There's no need to set the file modification date on item set, it's u…

garrettmoon:
…pdated automatically by the system. But we should be setting the date internally.

Improves the setObject method by ~5% on an iPhone 4.

### DIFF
--- a/PINCache/PINDiskCache.m
+++ b/PINCache/PINDiskCache.m
@@ -810,8 +810,6 @@ typedef NS_ENUM(NSUInteger, PINDiskCacheCondition) {
 
 - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key fileURL:(NSURL **)outFileURL
 {
-    NSDate *now = [[NSDate alloc] init];
-    
     if (!key || !object)
         return;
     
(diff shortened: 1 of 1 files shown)


PINCache/PINDiskCache.m
```diff
@@ -842,11 +840,8 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key fileURL:(NSURL **
         PINDiskCacheError(writeError);
         
         if (written) {
-            //Attempting to do this asynchronously seems to result in *worse* performance.
-            [self _locked_setFileModificationDate:now forURL:fileURL];
-            
             NSError *error = nil;
-            NSDictionary *values = [fileURL resourceValuesForKeys:@[ NSURLTotalFileAllocatedSizeKey ] error:&error];
+            NSDictionary *values = [fileURL resourceValuesForKeys:@[ NSURLContentModificationDateKey, NSURLTotalFileAllocatedSizeKey ] error:&error];
```
appleguy: @garrettmoon Great change!  Curious though, what is the reason to request this value?  Couldn't we just use -now, if the function is to update an LRU data structure?

garrettmoon:
My thought was correctness (I measured a difference of one second which is who cares) and the assumption it would be cheap because we're already getting the actual disk size already at the same time.

Separately, I thought about removing the actual disk cost as an option (instead using data length which should only be off by whatever the file system's block size or whatever it's called) in which case pulling the on disk modification date likely isn't worth it…

Anyway, might be worth adding an option for skipping all the reads after set :D

appleguy:
Useful context, all that makes sense - thanks!

Yeah, these do seem like legitimate points of engineering rigor.  My 2c is that, since this is a cache rather than a syncing service or such, the speed of avoiding further disk access contention is an asset worth more than the ~5% error in space and time management.

I agree it makes sense to provide as an option (maybe even default-off?), since there could certainly be use cases where highly predictable / exact behavior is expected.

		-- Scott

PR #119, opened 2016-10-06
Add ability to set operation priority

garrettmoon:


PINCache/PINOperationQueue.h
```diff
@@ -23,6 +23,7 @@ typedef NS_ENUM(NSUInteger, PINOperationQueuePriority) {
 - (instancetype)initWithMaxConcurrentOperations:(NSUInteger)maxConcurrentOperations concurrentQueue:(dispatch_queue_t)concurrentQueue NS_DESIGNATED_INITIALIZER;
 - (id <PINOperationReference>)addOperation:(dispatch_block_t)operation withPriority:(PINOperationQueuePriority)priority;
 - (void)cancelOperation:(id <PINOperationReference>)operationReference;
+- (void)setOperationPriority:(PINOperationQueuePriority)priority withReference:(id <PINOperationReference>)reference;
```
jparise: `forReference:` might be clearer here?

PINCache/PINOperationQueue.m
```diff
@@ -25,12 +25,14 @@ @interface PINOperationQueue () {
...
-  NSMutableOrderedSet *_allOperations;
+  NSMutableOrderedSet *_queuedOperations;
   NSMutableOrderedSet *_lowPriorityOperations;
   NSMutableOrderedSet *_defaultPriorityOperations;
   NSMutableOrderedSet *_highPriorityOperations;
   
-  NSHashTable *_canceledOperations;
+  NSMapTable *_referenceToOperations;
+  
+  NSArray <NSMutableOrderedSet *> *_prioritizedOperationQueues;
```
jparise: I might consider just using a C array here as a micro-optimization, sized to `PINOperationQueuePriorityHigh`. It doesn't need to retain any of its items because their lifetimes are already bound to this class' lifetime.

PINCache/PINOperationQueue.m
```diff
@@ -142,7 +133,27 @@ - (void)dealloc
 - (void)cancelOperation:(id <PINOperationReference>)operationReference
 {
   [self lock];
-    [_canceledOperations addObject:operationReference];
+    PINOperation *operation = [_referenceToOperations objectForKey:operationReference];
+    if (operation) {
+      for (NSMutableOrderedSet *queue in _prioritizedOperationQueues) {
```
maicki: Can we have some kind of connection operation -> queue so we don't have to iterate through all queues to remove the operation?

PINCache/PINOperationQueue.m
```diff
@@ -142,7 +133,27 @@ - (void)dealloc
...
+        [queue removeObject:operation];
```
jparise: Either 66% or 100% of these `removeObject:` calls are going to "fall". This should be a cheap operation though, right?

The alternative would be storing the priority on the operation itself and using that to look up its queue. That doesn't seem worth it unless this iteration is costly.

PINCache/PINOperationQueue.m
```diff
@@ -142,7 +133,27 @@ - (void)dealloc
...
+    }
+  [self unlock];
+}
+
+- (void)setOperationPriority:(PINOperationQueuePriority)priority withReference:(id <PINOperationReference>)operationReference
+{
+  [self lock];
+    PINOperation *operation = [_referenceToOperations objectForKey:operationReference];
+    if (operation) {
+      for (NSMutableOrderedSet *queue in _prioritizedOperationQueues) {
```
maicki: See comment above, so maybe we could remove this here too.

PINCache/PINOperationQueue.h
garrettmoon: withReference: seems to match existing API…

garrettmoon:
Alright, y'all win, I think you're right too ;)

PINCache/PINOperationQueue.h
```diff
@@ -12,6 +12,7 @@ typedef NS_ENUM(NSUInteger, PINOperationQueuePriority) {
   PINOperationQueuePriorityLow,
   PINOperationQueuePriorityDefault,
   PINOperationQueuePriorityHigh,
+  PINOperationQueuePriorityCount,
```
jparise: I don't think you actually use this in the latest version.

PR #121, opened 2016-10-08
Add waitUntilAllOperationsAreFinished to PINOperationQueue

maicki:
Adding `waitUntilAllOperationsAreFinished` to `PINOperationQueue`. Should work the same as `waitUntilAllOperationsAreFinished` of `NSOperationQueue`

One use case could be if the operation queue is used as a transaction.

### DIFF
--- a/PINCache/PINOperationQueue.h
+++ b/PINCache/PINOperationQueue.h
@@ -23,6 +23,16 @@ typedef NS_ENUM(NSUInteger, PINOperationQueuePriority) {
 - (instancetype)initWithMaxConcurrentOperations:(NSUInteger)maxConcurrentOperations concurrentQueue:(dispatch_queue_t)concurrentQueue NS_DESIGNATED_INITIALIZER;
 - (id <PINOperationReference>)addOperation:(dispatch_block_t)operation withPriority:(PINOperationQueuePriority)priority;
 - (void)cancelOperation:(id <PINOperationReference>)operationReference;
+
+/**
+ * Blocks the current thread until all of the receiver’s queued and executing operations finish executing.
+ *
+ * @discussion When called, this method blocks the current thread and waits for the receiver’s current and queued
+ * operations to finish executing. While the current thread is blocked, the receiver continues to launch already
+ * queued operations and monitor those that are executing.
+ */
+- (void)waitUntilAllOperationsAreFinished;
+
 - (void)setOperationPriority:(PINOperationQueuePriority)priority withReference:(id <PINOperationReference>)reference;
 
 @end
--- a/PINCache/PINOperationQueue.h
+++ b/PINCache/PINOperationQueue.h
@@ -30,6 +30,9 @@ typedef NS_ENUM(NSUInteger, PINOperationQueuePriority) {
  * @discussion When called, this method blocks the current thread and waits for the receiver’s current and queued
  * operations to finish executing. While the current thread is blocked, the receiver continues to launch already
  * queued operations and monitor those that are executing.
+ *
+ * @warning This should never be called from within an operation submitted to the PINOperationQueue as this will result
+ * in a deadlock.
  */
 - (void)waitUntilAllOperationsAreFinished;
 
--- a/PINCache/PINOperationQueue.m
+++ b/PINCache/PINOperationQueue.m
@@ -141,7 +141,7 @@
       NSMutableOrderedSet *queue = [self operationQueueWithPriority:operation.priority];
       [queue removeObject:operation];
       [_queuedOperations removeObject:operation];
-      dispatch_group_leave(_group); // Add operation to queue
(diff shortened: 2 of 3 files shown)


garrettmoon:
@maicki thanks for putting this and #122 up!

I'm wondering if these methods should actually be built in a transaction or group abstraction which utilizes a PINOperationQueue? That way you can limit the number of threads with one PINOperationQueue for multiple groups / transactions. If we build this functionality on PINOperationQueue, we will only be able to have thread limits per group / transaction *not* an entire system at once (i.e. ASDK, PINCache or PINRemoteImage).

It will also be unsafe to build something like a sharedOperationQueue without crazy side effects when someone decides to cancel all operations :)

garrettmoon:
Talked with @maicki on slack and his intention is to mirror NSOperationQueue which I think is a good idea. If we do add a sharedOperationQueue, we can assert on calling cancel all or waitUntilAll on just the sharedOperationQueue.

tests/PINCacheTests/PINOperationQueueTests.m
```diff
@@ -58,6 +58,44 @@ - (void)testAllOperationsRun
...
+{
+  const NSUInteger operationCount = 100;
+  __block NSInteger runnedOperations = 0;;
+  for (NSUInteger count = 0; count < operationCount; count++) {
+    [self.queue addOperation:^{
+      runnedOperations += 1;
+    } withPriority:PINOperationQueuePriorityDefault];
+  }
+  
+  //
```
garrettmoon: nit: remove or add comment

tests/PINCacheTests/PINOperationQueueTests.m
```diff
@@ -58,6 +58,44 @@ - (void)testAllOperationsRun
   XCTAssert(success == 0, @"Timed out before completing 100 operations");
 }
 
+- (void)testWaitUntilAllOperationsFinished
+{
+  const NSUInteger operationCount = 100;
+  __block NSInteger runnedOperations = 0;;
```
garrettmoon: `operationsRun` ?

tests/PINCacheTests/PINOperationQueueTests.m
```diff
@@ -58,6 +58,44 @@ - (void)testAllOperationsRun
...
+    [self.queue addOperation:^{
+      __strong PINOperationQueueTests *strongSelf = weakSelf;
+      runnedOperations += 1;
+      [strongSelf.queue addOperation:^{
+          runnedOperations += 1;
+      } withPriority:PINOperationQueuePriorityHigh];
+    } withPriority:PINOperationQueuePriorityDefault];
+  }
+
+  //
```
garrettmoon: nit: remove or add comment.

maicki:
@garrettmoon Comments addressed

PINCache/PINOperationQueue.h
```diff
@@ -23,6 +23,16 @@ typedef NS_ENUM(NSUInteger, PINOperationQueuePriority) {
 - (instancetype)initWithMaxConcurrentOperations:(NSUInteger)maxConcurrentOperations concurrentQueue:(dispatch_queue_t)concurrentQueue NS_DESIGNATED_INITIALIZER;
 - (id <PINOperationReference>)addOperation:(dispatch_block_t)operation withPriority:(PINOperationQueuePriority)priority;
 - (void)cancelOperation:(id <PINOperationReference>)operationReference;
+
+/**
+ * Blocks the current thread until all of the receiver’s queued and executing operations finish executing.
+ *
+ * @discussion When called, this method blocks the current thread and waits for the receiver’s current and queued
+ * operations to finish executing. While the current thread is blocked, the receiver continues to launch already
+ * queued operations and monitor those that are executing.
```
garrettmoon: Should we add @warning This should never be called from within an operation submitted to the PINOperationQueue as this will result in a deadlock.

PINCache/PINOperationQueue.m
```diff
@@ -136,6 +141,7 @@ - (void)cancelOperation:(id <PINOperationReference>)operationReference
       NSMutableOrderedSet *queue = [self operationQueueWithPriority:operation.priority];
       [queue removeObject:operation];
       [_queuedOperations removeObject:operation];
+      dispatch_group_leave(_group); // Add operation to queue
```
garrettmoon: Can you make the comment something like "Matching group_enter in addOperation:withPriority"

maicki: I removed the comment actually. Don't think we need one here.

PINCache/PINOperationQueue.h
maicki: Despite `NSOperationQueue` has the same problem and does not have any of that comment I think it can be valuable to have one. So I added the comment

PR #122, opened 2016-10-08
Add cancelAllOperations to PINOperationQueue

maicki:
Add support to cancel all operations. Furthermore created a new method for cancelling an operation that the lock needs to be hold. This used by `cancelOperation` as well as `cancelAllOperations` now.

### DIFF
--- a/PINCache/PINOperationQueue.m
+++ b/PINCache/PINOperationQueue.m
@@ -131,9 +131,9 @@
 - (void)cancelAllOperations
 {
   [self lock];
-  for (PINOperation *operation in [_referenceToOperations objectEnumerator]) {
-    [self locked_cancelOperation:operation.reference];
-  }
+    for (PINOperation *operation in [[_referenceToOperations copy] objectEnumerator]) {
+      [self locked_cancelOperation:operation.reference];
+    }
   [self unlock];
 }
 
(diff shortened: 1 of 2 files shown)


PINCache/PINOperationQueue.m
```diff
@@ -128,18 +128,33 @@ - (void)dealloc
   return reference;
 }
 
+- (void)cancelAllOperations
+{
+  [self lock];
+  for (PINOperation *operation in [_referenceToOperations objectEnumerator]) {
```
garrettmoon: Another advantage of the _referenceToOperations weakTable :) We wouldn't have been able to enumerate over _queuedOperations without making a copy. Do you know if we need a copy here? Is it safe to enumerate over it even as objects are potentially being nil'd out by locked_cancelOperation?

PINCache/PINOperationQueue.m
```diff
@@ -128,18 +128,33 @@ - (void)dealloc
   return reference;
 }
 
+- (void)cancelAllOperations
+{
+  [self lock];
```
garrettmoon: nit: can you indent the code within the lock?

maicki: Done

maicki:
@garrettmoon Addressed comments

PR #127, opened 2016-10-19
Adopt pin operation queue

garrettmoon:


PINCache/PINCache.m
```diff
@@ -110,7 +100,7 @@ - (void)objectForKey:(NSString *)key block:(PINCacheObjectBlock)block
     
     __weak PINCache *weakSelf = self;
     
-    dispatch_async(_concurrentQueue, ^{
+    [self.operationQueue addOperation:^{
```
appleguy: @garrettmoon minor, but might as well use _operationQueue

PINCache/PINCache.m
```diff
@@ -152,157 +141,85 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key block:(PINCacheOb
...
-        diskBlock = ^(PINDiskCache *diskCache, NSString *diskCacheKey, id <NSCoding> memoryCacheObject) {
-            dispatch_group_leave(group);
-        };
-    }
+    PINOperationGroup *group = [PINOperationGroup asyncOperationGroupWithQueue:self.operationQueue];
     
-    [_memoryCache removeObjectForKey:key block:memBlock];
-    [_diskCache removeObjectForKey:key block:diskBlock];
+    [group addOperation:^{
+        [_memoryCache removeObjectForKey:key];
```
appleguy: Is it bad that this captures self by using instance variable?  This seems maybe OK because the operations should have finite lifecycle.  If it is OK, we should remove the weak / strongSelf below in the completion, since it wouldn't be worth using a weak variable if everything but the completion block will keep self alive.

PINCache/PINCache.m
```diff
@@ -152,157 +141,85 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key block:(PINCacheOb
...
-        
-        memBlock = ^(PINMemoryCache *cache) {
-            dispatch_group_leave(group);
-        };
-        
-        diskBlock = ^(PINDiskCache *cache) {
-            dispatch_group_leave(group);
-        };
-    }
+    PINOperationGroup *group = [PINOperationGroup asyncOperationGroupWithQueue:self.operationQueue];
```
appleguy: In all these cases, could use _operationQueue; same question on strongSelf

PINCache/PINCache.m
```diff
@@ -152,157 +141,85 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key block:(PINCacheOb
...
-        };
-        
-        diskBlock = ^(PINDiskCache *diskCache, NSString *diskCacheKey, id <NSCoding> memoryCacheObject) {
-            dispatch_group_leave(group);
-        };
-    }
-    
-    [_memoryCache setObject:object forKey:key block:memBlock];
-    [_diskCache setObject:object forKey:key block:diskBlock];
+    [group addOperation:^{
```
garrettmoon: The idea is that they should be parallelized. I also want to be able cancel them (in the future) and separating allows that to be more likely to occur.

Adlai-Holler:
Wooooo! ⚡️

Issue #134, opened 2016-10-27
Assertion Failure on Single-Core Devices


Adlai-Holler:
Launching an app that uses PINCache on an iPhone 4 causes immediate assertion failure: `Max concurrent operations must be greater than 1. If it's one, just use a serial queue!`

Best bet is to remove this assertion – it feels over-eager IMO.

garrettmoon:
I'm not positive it will *work* with a maxConcurrentOperations of 1. It might, but I'll need to test. I think the issue is this line:
`[[PINOperationQueue alloc] initWithMaxConcurrentOperations:[[NSProcessInfo processInfo] activeProcessorCount]];`

which should be
`[[PINOperationQueue alloc] initWithMaxConcurrentOperations:[[NSProcessInfo processInfo] activeProcessorCount] * 2];

garrettmoon:
@Adlai-Holler Happy to discuss removing the assertion, but I think it actually catches a legitimate issue.

PR #139, opened 2016-11-02
Have sharedOperationQueue have at least 2 max operations.

garrettmoon:
On devices without multiple cores it's still useful to allow multiple
operations to run at once.

Hopefully addresses #134 134


appleguy:
I would suggest a minimum of 2 * the core count.  I suspect we'd find that a higher limit is actually worthwhile too, due to various filesystem implementation complexities (blocked reads, flushing interval for writes).

garrettmoon:
Since there isn't much use of the sharedQueue yet, let's leave this as is until we know more. But I suspect you're right.

PR #142, opened 2016-11-13
Optimize trash emptying in disk cache

nguyenhuy:
Disk cache now removes its trash directory as a whole instead of removing files inside one by one #136, #137.

PR #141, opened 2016-11-11
Support operation coallescing

nguyenhuy:
Add a generic operation coallescing mechanism and use it to fix #135, #137.

Issue #137, opened 2016-10-28
Reduce frequency / cost of calls to +emptyTrash and -trimToByteSize: (tolerance % for cleanup)


appleguy:
Currently we spend a large total resource on calls to these methods.

The example here is from resuming the app on iPod 5G _without_ any user interaction.  Note that when this happens, we re-render a significant number of views that expand their working ranges, so accesses to the cache are common; however, we shouldn't be downloading any new images.

The file read operations would be expected to be the most expensive part of this (in addition to the decodes).  However the total cache access / file read costs are very modest in comparison to the cleanup operations, which are a large multiple of the read costs!

Adding the two type of dispatch stacks, there's about 2,000ms of work in total.  Only 360ms are file reads!

<img width="850" alt="screen shot 2016-10-28 at 4 29 55 pm" src="https://cloud.githubusercontent.com/assets/565251/19825318/57953554-9d2c-11e6-85c9-51c5dd2ed401.png">
<img width="857" alt="screen shot 2016-10-28 at 4 31 10 pm" src="https://cloud.githubusercontent.com/assets/565251/19825284/f9dedcb2-9d2b-11e6-9728-020a2b8d2e91.png">

nguyenhuy:
Fixed!

Issue #136, opened 2016-10-28
+emptyTrash operation can be optimized by deleting directory as a whole (rather than enumerating files).


appleguy:
I tried this locally and it's a great win - but a little bit of extra work is needed to make it safe.  We simply need to ensure another directory is created to replace it, before any other file is moved there.

A great implementation would create the trash in a random place, and then when the trash is emptied, delete that place + have the sharedTrashURL's cached URL be cleared.  Then the next time sharedTrashURL is called, we would see there's no cached global URL, and create a new UUID-named trash folder.  This would eliminate any checks for whether the directory exists too.

This trace was taken on an iPod 5G, after suspending the app, measuring _only the app resume_ operation - with no user interaction.

<img width="847" alt="screen shot 2016-10-28 at 3 21 06 pm" src="https://cloud.githubusercontent.com/assets/565251/19824190/3c659242-9d22-11e6-9322-c7f46f9700c8.png">

nguyenhuy:
Optimization merged!

PR #146, opened 2016-11-18
Address comments and improve PINCacheTests

nguyenhuy:
Follow up on #142 which was accidentally merged together with #141. 

cc #136, #137.


garrettmoon:
Thanks for following up @nguyenhuy :)

PR #154, opened 2016-12-20
Add a build script

garrettmoon:


garrettmoon:
All the tutorials I'm seeing focus on compilation, build is probably a misnomer in this case, it really should be called tests.sh (will happily do this). Does it still make sense to use a makefile? Do you know where to find documentation that focuses on this aspect of a makefile? Everything I'm seeing focuses on the compilation aspect, i.e. here's a set of source files you need to compile for this step.

jparise:
It would look something like this (totally untested):

```make
PLATFORM="platform=iOS Simulator,OS=10.1,name=iPhone 7"
SDK="iphonesimulator10.1"

.PHONY: all lint test carthage

lint:
    pod lib lint --allow-warnings

test:
     xcodebuild clean test -destination "$PLATFORM" -sdk "$SDK" -project tests/PINCache.xcodeproj -scheme PINCacheTests ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO | xcpretty -t; test ${PIPESTATUS[0]} -eq 0

carthage:
    carthage build --no-skip-current

all: lint test carthage
```

... and then you would invoke different targets as `make lint` or `make lint test` or `make all`.

PR #156, opened 2016-12-31
Buck improvements

nguyenhuy:
- Merge PINCacheCore, PINCache and PINCacheMacOS targets into a single target.
- Compile PINDishCache.m with -fobjc-arc-exceptions flag.
- Move test targets to a separate BUCK file. The main BUCK file can be re-used by users in their own projects after dependency resolution (i.e via CocoaPods).
- Remove unnecessary preprocessor flags.
- Now that we have multiple BUCK files, parse them in parallel.
- Specify default xctool destination.


garrettmoon:
Thanks @nguyenhuy !

PR #160, opened 2017-01-11
Migrate to External PINOperation

Adlai-Holler:
PINOperation now has its own home at https://github.com/pinterest/PINOperation and so PINCache needs to bring it in as a dependency.

This also renames `Coallescing` to `Coalescing`, so we can get the name changed before the release.


garrettmoon:
Last failure:
 -> PINCache (3.0.1-beta.2)
    - ERROR | [PINCache/Core,PINCache/Arc-exception-safe] file patterns: The `source_files` pattern did not match any file.

Adlai-Holler:
@garrettmoon I think the spec lint is failing because it's using the _new_ spec but the _old_ code – that is, it pulls down the `3.0.1-beta.2` tag and tries to validate this spec against that version of the repo, which won't work.

garrettmoon:
It looks like the build_tests.sh script no longer works. I'm not sure this is valuable at this point anyway. Mind removing?

garrettmoon:
Thanks so much for the cleanup @Adlai-Holler !

PR #162, opened 2017-01-24
Update memory cache header

ToroLiu:
Some `nullable` will make trouble. For example, `objectForKey` with no block parameter. In swift, it's trouble.

For example, the following are ambiguous in Swift3. They have the same method signature.

```
  - (void)objectForKey:(NSString *)key block:(nullable PINMemoryCacheObjectBlock)block;
  - (nullable id)objectForKey:(NSString *)key
```

### DIFF
--- a/Source/PINMemoryCache.h
+++ b/Source/PINMemoryCache.h
@@ -169,7 +169,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param key The key associated with the object.
  @param block A block to be executed concurrently after the containment check happened
  */
-- (void)containsObjectForKey:(NSString *)key block:(PINMemoryCacheContainmentBlock)block;
+- (void)containsObjectForKey:(NSString *)key block:(null_unspecified PINMemoryCacheContainmentBlock)block;
 
 /**
  Retrieves the object for the specified key. This method returns immediately and executes the passed
@@ -178,7 +178,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param key The key associated with the requested object.
  @param block A block to be executed concurrently when the object is available.
  */
-- (void)objectForKey:(NSString *)key block:(PINMemoryCacheObjectBlock)block;
+- (void)objectForKey:(NSString *)key block:(null_unspecified PINMemoryCacheObjectBlock)block;
 
 /**
  Stores an object in the cache for the specified key. This method returns immediately and executes the
@@ -188,7 +188,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param key A key to associate with the object. This string will be copied.
  @param block A block to be executed concurrently after the object has been stored, or nil.
  */
-- (void)setObject:(id)object forKey:(NSString *)key block:(PINMemoryCacheObjectBlock)block;
+- (void)setObject:(id)object forKey:(NSString *)key block:(null_unspecified PINMemoryCacheObjectBlock)block;
 
 /**
  Stores an object in the cache for the specified key and the specified cost. If the cost causes the total
@@ -201,7 +201,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param cost An amount to add to the <totalCost>.
  @param block A block to be executed concurrently after the object has been stored, or nil.
  */
-- (void)setObject:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost block:(PINMemoryCacheObjectBlock)block;
+- (void)setObject:(id)object forKey:(NSString *)key withCost:(NSUInteger)cost block:(null_unspecified PINMemoryCacheObjectBlock)block;
 
 /**
  Removes the object for the specified key. This method returns immediately and executes the passed
@@ -210,7 +210,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param key The key associated with the object to be removed.
  @param block A block to be executed concurrently after the object has been removed, or nil.
  */
-- (void)removeObjectForKey:(NSString *)key block:(PINMemoryCacheObjectBlock)block;
+- (void)removeObjectForKey:(NSString *)key block:(null_unspecified PINMemoryCacheObjectBlock)block;
 
 /**
  Removes all objects from the cache that have not been used since the specified date.
@@ -220,7 +220,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param date Objects that haven't been accessed since this date are removed from the cache.
  @param block A block to be executed concurrently after the cache has been trimmed, or nil.
  */
-- (void)trimToDate:(NSDate *)date block:(PINMemoryCacheBlock)block;
+- (void)trimToDate:(NSDate *)date block:(null_unspecified PINMemoryCacheBlock)block;
 
 /**
  Removes objects from the cache, costliest objects first, until the <totalCost> is below the specified
@@ -230,7 +230,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param cost The total accumulation allowed to remain after the cache has been trimmed.
  @param block A block to be executed concurrently after the cache has been trimmed, or nil.
  */
-- (void)trimToCost:(NSUInteger)cost block:(PINMemoryCacheBlock)block;
+- (void)trimToCost:(NSUInteger)cost block:(null_unspecified PINMemoryCacheBlock)block;
 
 /**
  Removes objects from the cache, ordered by date (least recently used first), until the <totalCost> is below
@@ -240,7 +240,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  @param cost The total accumulation allowed to remain after the cache has been trimmed.
  @param block A block to be executed concurrently after the cache has been trimmed, or nil.
  */
-- (void)trimToCostByDate:(NSUInteger)cost block:(PINMemoryCacheBlock)block;
+- (void)trimToCostByDate:(NSUInteger)cost block:(null_unspecified PINMemoryCacheBlock)block;
 
 /**
  Removes all objects from the cache. This method returns immediately and executes the passed block after
@@ -248,7 +248,7 @@ typedef void (^PINMemoryCacheContainmentBlock)(BOOL containsObject);
  
  @param block A block to be executed concurrently after the cache has been cleared, or nil.
  */
-- (void)removeAllObjects:(PINMemoryCacheBlock)block;
+- (void)removeAllObjects:(null_unspecified PINMemoryCacheBlock)block;
 
 /**
  Loops through all objects in the cache with reads and writes suspended. Calling serial methods which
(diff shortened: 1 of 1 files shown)


garrettmoon:
Hi @ToroLiu is there another way to fix this issue? It's important to be able to call the objectForKey:block: with a nil block because it's the asynchronous version (i.e. it doesn't block the thread while waiting to get the object) of the method and objectForKey: is synchronous (i.e. it blocks the thread until it gets the object).

ToroLiu:
In Swift3, they are with the same method signatures. The `__nullable` attribute has meaning in Swift.

And, @garrettmoon. If the block is nil, the method `objectForKey:block:` return `Void`. How do you get the return object with `asynchronous` way?
```
// Original
open class PINMemoryCache : NSObject, PINCacheObjectSubscripting {
  // ...

  open func object(forKey key: String, block: PINCache.PINMemoryCacheObjectBlock = nil)

  // Due to the above method's block is nil by default. It will has the same method signature with this.
  // So, Swift compiler don't know which one is correct.
  open func object(forKey key: String?) -> Any?

}
```

ToroLiu:
What I modify is the header of `PMMemoryCache`. It's no problem with `PINCache`.

garrettmoon:
I understand the issue, it sounds like a real problem! I'm just questioning the solution (I'm no Swift expert though).

In the case of `objectForKey:block:`, there's no reason to call it with a nil block except perhaps for side effects. However, with something like `setObjectForKey:block:` there is, you may not care when the object is set, so providing a nil block allows you to ignore the results without the overhead of PINMemoryCache needing to call the block.

Again, I see the problem, I'm just wondering if there's another solution than eliminating the nullable attribute.

Separately, isn't this an issue with PINCache and PINDiskCache as well? If not, do you know why not?

ToroLiu:
Maybe the `PINCache.h` and `PINMemoryCache.h` should be review again about this problem. 

The `nullable` has meaning in `Swift`, but in Obj-C, maybe it has no meaning ? 


```
// In PINCache.h
- (void)setObject:(id <NSCoding>)object forKey:(NSString *)key block:(nullable PINCacheObjectBlock)block;

- (void)objectForKey:(NSString *)key block:(PINCacheObjectBlock)block;
```

The following is PINCache in Swift. Converted by Xcode.
```
import UIKit

//  PINCache is a modified version of TMCache
//  Modifications by Garrett Moon
//  Copyright (c) 2015 Pinterest. All rights reserved.

/**
 A callback block which provides only the cache as an argument
 */
public typealias PINCacheBlock = (PINCache) -> Swift.Void

/**
 A callback block which provides the cache, key and object as arguments
 */
public typealias PINCacheObjectBlock = (PINCache, String, Any?) -> Swift.Void

/**
 A callback block which provides a BOOL value as argument
 */
public typealias PINCacheObjectContainmentBlock = (Bool) -> Swift.Void


/**
 `PINCache` is a thread safe key/value store designed for persisting temporary objects that are expensive to
 reproduce, such as downloaded data or the results of slow processing. It is comprised of two self-similar
 stores, one in memory (<PINMemoryCache>) and one on disk (<PINDiskCache>).
 
 `PINCache` itself actually does very little; its main function is providing a front end for a common use case:
 a small, fast memory cache that asynchronously persists itself to a large, slow disk cache. When objects are
 removed from the memory cache in response to an "apocalyptic" event they remain in the disk cache and are
 repopulated in memory the next time they are accessed. `PINCache` also does the tedious work of creating a
 dispatch group to wait for both caches to finish their operations without blocking each other.
 
 The parallel caches are accessible as public properties (<memoryCache> and <diskCache>) and can be manipulated
 separately if necessary. See the docs for <PINMemoryCache> and <PINDiskCache> for more details.

 @warning when using in extension or watch extension, define PIN_APP_EXTENSIONS=1
 */
open class PINCache : NSObject, PINCacheObjectSubscripting {

    
    /// @name Core
    
    /**
     The name of this cache, used to create the <diskCache> and also appearing in stack traces.
     */
    open var name: String { get }

    
    /**
     Synchronously retrieves the total byte count of the <diskCache> on the shared disk queue.
     */
    open var diskByteCount: UInt { get }

    
    /**
     The underlying disk cache, see <PINDiskCache> for additional configuration and trimming options.
     */
    open var diskCache: PINDiskCache { get }

    
    /**
     The underlying memory cache, see <PINMemoryCache> for additional configuration and trimming options.
     */
    open var memoryCache: PINMemoryCache { get }

    
    /// @name Initialization
    
    /**
     A shared cache.
     
     @result The shared singleton cache instance.
     */
    open class func shared() -> Self

    
    /**
     Multiple instances with the same name are allowed and can safely access
     the same data on disk thanks to the magic of seriality. Also used to create the <diskCache>.
     
     @see name
     @param name The name of the cache.
     @result A new cache with the specified name.
     */
    public convenience init(name: String)

    
    /**
     Multiple instances with the same name are allowed and can safely access
     the same data on disk thanks to the magic of seriality. Also used to create the <diskCache>.
     
     @see name
     @param name The name of the cache.
     @param fileExtension The file extension for files on disk.
     @result A new cache with the specified name.
     */
    public convenience init(name: String, fileExtension: String?)

    
    /**
     Multiple instances with the same name are allowed and can safely access
     the same data on disk thanks to the magic of seriality. Also used to create the <diskCache>.
     
     @see name
     @param name The name of the cache.
     @param rootPath The path of the cache on disk.
     @param fileExtension The file extension for files on disk.
     @result A new cache with the specified name.
     */
    public convenience init(name: String, rootPath: String, fileExtension: String?)

    
    /**
     Multiple instances with the same name are allowed and can safely access
     the same data on disk thanks to the magic of seriality. Also used to create the <diskCache>.
     Initializer allows you to override default NSKeyedArchiver/NSKeyedUnarchiver serialization for <diskCache>.
     You must provide both serializer and deserializer, or opt-out to default implementation providing nil values.
     
     @see name
     @param name The name of the cache.
     @param rootPath The path of the cache on disk.
     @param serializer   A block used to serialize object before writing to disk. If nil provided, default NSKeyedArchiver serialized will be used.
     @param deserializer A block used to deserialize object read from disk. If nil provided, default NSKeyedUnarchiver serialized will be used.
     @param fileExtension The file extension for files on disk.
     @result A new cache with the specified name.
     */
    public init(name: String, rootPath: String, serializer: PINCache.PINDiskCacheSerializerBlock?, deserializer: PINCache.PINDiskCacheDeserializerBlock?, fileExtension: String?)

    
    /// @name Asynchronous Methods
    
    /**
     This method determines whether an object is present for the given key in the cache. This method returns immediately
     and executes the passed block after the object is available, potentially in parallel with other blocks on the
     <concurrentQueue>.
     
     @see containsObjectForKey:
     @param key The key associated with the object.
     @param block A block to be executed concurrently after the containment check happened
     */
    open func containsObject(forKey key: String, block: PINCache.PINCacheObjectContainmentBlock)

    
    /**
     Retrieves the object for the specified key. This method returns immediately and executes the passed
     block after the object is available, potentially in parallel with other blocks on the <concurrentQueue>.
     
     @param key The key associated with the requested object.
     @param block A block to be executed concurrently when the object is available.
     */
    open func object(forKey key: String, block: PINCache.PINCacheObjectBlock)

    
    /**
     Stores an object in the cache for the specified key. This method returns immediately and executes the
     passed block after the object has been stored, potentially in parallel with other blocks on the <concurrentQueue>.
     
     @param object An object to store in the cache.
     @param key A key to associate with the object. This string will be copied.
     @param block A block to be executed concurrently after the object has been stored, or nil.
     */
    open func setObject(_ object: NSCoding, forKey key: String, block: PINCache.PINCacheObjectBlock? = nil)

    
    /**
     Stores an object in the cache for the specified key and the specified memory cost. If the cost causes the total
     to go over the <memoryCache.costLimit> the cache is trimmed (oldest objects first). This method returns immediately
     and executes the passed block after the object has been stored, potentially in parallel with other blocks
     on the <concurrentQueue>.
     
     @param object An object to store in the cache.
     @param key A key to associate with the object. This string will be copied.
     @param cost An amount to add to the <memoryCache.totalCost>.
     @param block A block to be executed concurrently after the object has been stored, or nil.
     */
    open func setObject(_ object: NSCoding, forKey key: String, withCost cost: UInt, block: PINCache.PINCacheObjectBlock? = nil)

    
    /**
     Removes the object for the specified key. This method returns immediately and executes the passed
     block after the object has been removed, potentially in parallel with other blocks on the <concurrentQueue>.
     
     @param key The key associated with the object to be removed.
     @param block A block to be executed concurrently after the object has been removed, or nil.
     */
    open func removeObject(forKey key: String, block: PINCache.PINCacheObjectBlock? = nil)

    
    /**
     Removes all objects from the cache that have not been used since the specified date. This method returns immediately and
     executes the passed block after the cache has been trimmed, potentially in parallel with other blocks on the <concurrentQueue>.
     
     @param date Objects that haven't been accessed since this date are removed from the cache.
     @param block A block to be executed concurrently after the cache has been trimmed, or nil.
     */
    open func trim(to date: Date, block: PINCache.PINCacheBlock? = nil)

    
    /**
     Removes all objects from the cache.This method returns immediately and executes the passed block after the
     cache has been cleared, potentially in parallel with other blocks on the <concurrentQueue>.
     
     @param block A block to be executed concurrently after the cache has been cleared, or nil.
     */
    open func removeAllObjects(_ block: PINCache.PINCacheBlock? = nil)

    
    /// @name Synchronous Methods
    
    /**
     This method determines whether an object is present for the given key in the cache.
     
     @see containsObjectForKey:block:
     @param key The key associated with the object.
     @result YES if an object is present for the given key in the cache, otherwise NO.
     */
    open func containsObject(forKey key: String) -> Bool

    
    /**
     Retrieves the object for the specified key. This method blocks the calling thread until the object is available.
     Uses a lock to achieve synchronicity on the disk cache.
     
     @see objectForKey:block:
     @param key The key associated with the object.
     @result The object for the specified key.
     */
    open func object(forKey key: String) -> Any?

    
    /**
     Stores an object in the cache for the specified key. This method blocks the calling thread until the object has been set.
     Uses a lock to achieve synchronicity on the disk cache.
     
     @see setObject:forKey:block:
     @param object An object to store in the cache.
     @param key A key to associate with the object. This string will be copied.
     */
    open func setObject(_ object: NSCoding, forKey key: String)

    
    /**
     Stores an object in the cache for the specified key and the specified memory cost. If the cost causes the total
     to go over the <memoryCache.costLimit> the cache is trimmed (oldest objects first). This method blocks the calling thread
     until the object has been stored.
     
     @param object An object to store in the cache.
     @param key A key to associate with the object. This string will be copied.
     @param cost An amount to add to the <memoryCache.totalCost>.
     */
    open func setObject(_ object: NSCoding, forKey key: String, withCost cost: UInt)

    
    /**
     Removes the object for the specified key. This method blocks the calling thread until the object
     has been removed.
     Uses a lock to achieve synchronicity on the disk cache.
     
     @param key The key associated with the object to be removed.
     */
    open func removeObject(forKey key: String)

    
    /**
     Removes all objects from the cache that have not been used since the specified date.
     This method blocks the calling thread until the cache has been trimmed.
     Uses a lock to achieve synchronicity on the disk cache.
     
     @param date Objects that haven't been accessed since this date are removed from the cache.
     */
    open func trim(to date: Date)

    
    /**
     Removes all objects from the cache. This method blocks the calling thread until the cache has been cleared.
     Uses a lock to achieve synchronicity on the disk cache.
     */
    open func removeAllObjects()
}

```

ToroLiu:
Maybe the `nullable` in the method parameter should be `null_unspecified `.

Refer link: http://stackoverflow.com/a/39970305/419348

Adlai-Holler:
@ToroLiu Note: I did a pass on this header in #164, you may want to give feedback there.

garrettmoon:
@ToroLiu We've made some extensive changes to the API in order to address this issue. I'm going to close this pull request, but please comment back if you don't feel like we addressed it!

Issue #143, opened 2016-11-13
Make Full Pathname configurable


docterd:
Currently PINCache stores always it's data under com.pinterest.PINDiskCache.name. This is very confusing for the user under macOS since he expect his cached data unter the bundler identifier within the Cache folder.

hannahmbanana:
@docterd - we'd be happy to accept a pull request if you'd like to add this functionality!

PR #163, opened 2017-01-31
Expose cache URL class method and allow specification of name prefix

garrettmoon:
This exposes a method which given the parameters you would pass into
PINDiskCache would allow you to know the URL the disk cache will end
up being at.

It also allows setting the cache name prefix so it can be your application
identifier instead of com.pinterest.PINDiskCache.(name)

### DIFF
--- a/Source/PINDiskCache.h
+++ b/Source/PINDiskCache.h
@@ -80,12 +80,26 @@ typedef id<NSCoding> __nonnull(^PINDiskCacheDeserializerBlock)(NSData* data, NSS
 
 
 #pragma mark -
+
+/**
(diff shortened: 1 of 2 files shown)


garrettmoon:
This should fix #143

garrettmoon:
So this is for the name of the folder. Currently it's named com.pinterest.PINDiskCache.(name), this allows you to name it com.yourcompany.yourapp.(name)

The extension is for the file backing in the actual cache, i.e. by default objects are stored as Key but you can also store them as key.arbitraryFileType if you wish.

Adlai-Holler:
Ah gotcha! 👍

PR #164, opened 2017-01-31
Update Objective-C, Prohibit Subclassing

Adlai-Holler:
Lately I've learned some fancy new Objective-C tricks!

- Set project indentation to use spaces.
- Prohibit subclassing of all three main classes. They're not suitable for subclassing, and this could possibly mean static dispatch one day. **Breaking API**
- Replace `__nullable` and `__nonnull` (Xcode 6 names) with `nullable` or `_Nullable` (Xcode 7 names).
- Mark more arguments nonnull. We may handle `nil` in production but nullability also communicates how something should be used. For example, `setObject:forKey:` now requires the object and key because it just doesn't make sense to pass `nil` here. **Breaking API for Swift**
- Replace `+(instancetype)sharedThing` with `@property (class, strong, readonly) PINThing *sharedThing`. **Breaking API for Swift**
- Mark blocks as `noescape` where appropriate. **API improvement for Swift**


Adlai-Holler:
@garrettmoon I'm not married to any of these changes, I just wanted to throw them all out and see what sticks, so let me know if you think it's too far.

Adlai-Holler:
Garrett, much like Jesus of Nazareth so long ago, sacrificed himself to redeem this PR in #169 .

PR #167, opened 2017-02-01
Introduce PINCaching protocol

maicki:
The PINCaching protocol contains methods that all three cache classes are implementing and all declared in the header. With PINCaching we can have a consistent API across all components and could mostly remove some duplicated API definitions in the header files too.

Furthermore with this PR we will rename the asynchronous API's to a more expressive name.

### TODOs
- [x] Rename async methods
- [x] Deprecate old async methods
- [x] Look over documentation and small cleanup due to async method deprecations
- [x] Rename async API to `setObjectAsync:forKey:completion:`

### DIFF
--- a/Source/PINCaching.h
+++ b/Source/PINCaching.h
@@ -118,7 +118,7 @@ typedef void (^PINCacheObjectContainmentBlock)(BOOL containsObject);
 /**
  This method determines whether an object is present for the given key in the cache.
  
- @see containsObjectForKey:block:
+ @see containsObjectForKeyAsync:completion:
  @param key The key associated with the object.
  @result YES if an object is present for the given key in the cache, otherwise NO.
  */
@@ -128,7 +128,7 @@ typedef void (^PINCacheObjectContainmentBlock)(BOOL containsObject);
  Retrieves the object for the specified key. This method blocks the calling thread until the object is available.
  Uses a lock to achieve synchronicity on the disk cache.
  
- @see objectForKey:block:
+ @see objectForKeyAsync:completion:
  @param key The key associated with the object.
  @result The object for the specified key.
  */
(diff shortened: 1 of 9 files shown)


maicki:
@garrettmoon I'm not pretty sure about the current async API. We could also go this way:
```
// Current
- (void)asyncSetObject:(id)object ...

// Alternative
- (void)setObjectAsync:(id)object ...
```

garrettmoon:
@maicki Yeah, now that I see it, I'm leaning towards setObjectAsync:

garrettmoon:
Yeah, let's go with `setObjectAsync:forKey:completion:`

maicki:
@garrettmoon Cool sounds good will change that.

maicki:
@garrettmoon Ready for review. Thanks!

Source/PINCaching.h
```diff
@@ -0,0 +1,185 @@
...
+ @param key The key associated with the object.
+ @result The object for the specified key.
+ */
+- (nullable id)objectForKey:(NSString *)key;
+
+/**
+ Stores an object in the cache for the specified key. This method blocks the calling thread until the object has been set.
+ Uses a lock to achieve synchronicity on the disk cache.
+ 
+ @see setObject:forKey:block:
```
garrettmoon: Can you update these comments to point to the new async versions? There appear to be a few more in this file.

Source/PINDiskCache.m
```diff
@@ -876,11 +880,20 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key
...
-- (void)setObject:(id)object forKeyedSubscript:(NSString *)key
+- (void)setObject:(id <NSCoding>)object forKey:(NSString *)key withCost:(NSUInteger)cost
 {
     [self setObject:object forKey:key];
 }
 
+- (void)setObject:(id)object forKeyedSubscript:(NSString *)key
+{
+    if (object == nil) {
+        [self removeObjectForKey:key];
```
garrettmoon: This is a behavior change, I assume it matches NSMutableDictionary?

maicki: Yep that's the same behavior as `NSMutableDictionary`

PR #173, opened 2017-02-16
Make Cartfile public for PINOperation dependency

zachwaugh:
Currently, using PINCache from PINRemoteImage is failing at run-time since PINCache depends on the PINOperation framework, and it's not getting built automatically due to the Cartfile being private. Simple change to make it public so anything that loads PINCache via Carthage will also build PINOperation.


garrettmoon:
Hey @zachwaugh I'm actually completely overhauling everything to better support Carthage *and* use it as the build system for the PIN* frameworks. Mind inputing on this PR: https://github.com/pinterest/PINCache/pull/171

garrettmoon:
And if you're interested in PINRemoteImage: https://github.com/pinterest/PINRemoteImage/pull/314

zachwaugh:
Thanks @garrettmoon, I see #171 was merged and makes the Cartfile public, so this is no longer necessary.

garrettmoon:
Thanks! Hopefully the new setup is better for Carthage, would love feedback if you have any.

PR #179, opened 2017-03-01
Add Swift examples to readmit and update Objective-C example to new async API

### LINKED ISSUE #178: How do you use this in Swift?
How can i use this lib that I know is super important in swift projects?
Are there any examples to reference?

thank you for this lib!

maicki:
Kind of resolves #178


garrettmoon:
Thank you :)

PR #181, opened 2017-03-20
Removes period from encoded characters

steam:
Encoding periods prevents mp4 files stored in `PINCache` from playing correctly in `AVFoundation`’s `AVPlayerItem` objects. To my knowledge it should be fine to leave periods unencoded.

### DIFF
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -231,7 +231,7 @@ static NSURL *_sharedTrashURL;
     }
     
     if ([string respondsToSelector:@selector(stringByAddingPercentEncodingWithAllowedCharacters:)]) {
-        NSString *encodedString = [string stringByAddingPercentEncodingWithAllowedCharacters:[[NSCharacterSet characterSetWithCharactersInString:@".:/%"] invertedSet]];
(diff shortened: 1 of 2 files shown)


steam:
Thanks @maicki, I hadn't considered existing data.

garrettmoon:
As this is a cache, I'm not *super* concerned about existing data being lost. This seems like good timing as well as we're barreling towards a major release.

garrettmoon:
Sorry, I'm going to revert this (very bad form, sorry, I should have thought through more before clicking the button): It's likely the users of this cache (i.e. PINRemoteImage) will have most items invalidated… I'm wondering if we can make some kind of migration tool?

garrettmoon:
Another solution would be to pull out the characters that are encoded into a separate method which could be overridden if necessary?

steam:
Thanks for the thoughtful consideration @garrettmoon. The overridable method approach is probably the easiest but it feels like an unnecessary expansion of the api since encoding the period character is not required for image caching to work. We were using PINRemoteImage and adding the mp4s into the same cache. We actually just ran into a hiccup with our mp4 approach and reverted to using FLAnimatedImage for standard gifs. I'll look into writing a migration into the PR when we reintroduce mp4s in the form of user submitted video.

garrettmoon:
Sweet, thank you @steam

PR #185, opened 2017-04-04
Revert "Removes period from encoded characters"

garrettmoon:
Reverts pinterest/PINCache#181


garrettmoon:
Sorry, need to revert https://github.com/pinterest/PINCache/pull/181 as I didn't fully think through the consequences.

Issue #195, opened 2017-09-01
deserialize failure fails silently


justin1121:
Hello, 

I noticed this piece of code fails silently if the deserialization fails on one of the objects. I think it should probably be a fatal and just print out the exception and crash. Spent a while debugging, trying to figure out what was going on.

https://github.com/pinterest/PINCache/blob/master/Source/PINDiskCache.m#L856

garrettmoon:
@justin1121 The history behind this is we were hearing reports of infrequent crashes. The theory was that somehow the objects on disk were corrupted, perhaps because the app crashed while serializing to disk. In that case, it's better to drop the object as this is a cache, not a guaranteed backing store than crash perhaps over and over again on startup. Seem reasonable?

garrettmoon:
Perhaps we should add an assert for debugging purposes? I'd be open to that! (Though the PINDiskError is meant to aid in debugging)

justin1121:
Yeah that makes sense. I think doing something like just printing PINDiskCacheError would help. The problem with the one that's there now is that it just prints the error that may come out of removing the file. If there's no error removing the file then it wouldn't print anything.

justin1121:
Check out that pull request. Think it will help with the problems I was having. Thanks!

garrettmoon:
Awesome, thanks for putting it up!

PR #196, opened 2017-09-05
#trivial print exception if deserialization fails

justin1121:
Fixes https://github.com/pinterest/PINCache/issues/195

### DIFF
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -16,9 +16,7 @@
 [[NSString stringWithUTF8String:__FILE__] lastPathComponent], \
 __LINE__, [error localizedDescription]); }
 
-#define PINDiskCacheException(exception) if (exception) { NSLog(@"%@ (%d) ERROR: %@", \
-[[NSString stringWithUTF8String:__FILE__] lastPathComponent], \
-__LINE__, [exception reason]); }
+#define PINDiskCacheException(exception) if (exception) { NSAssert(NO, [exception reason]); }
 
 NSString * const PINDiskCachePrefix = @"com.pinterest.PINDiskCache";
 static NSString * const PINDiskCacheSharedName = @"PINDiskCacheShared";
@@ -862,6 +860,7 @@ static NSURL *_sharedTrashURL;
                   [self lock];
                       [[NSFileManager defaultManager] removeItemAtPath:[fileURL path] error:&error];
                   [self unlock];
+                  PINDiskCacheError(error)
                   PINDiskCacheException(exception);
               }
               [self lock];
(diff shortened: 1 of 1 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
```diff
@@ -858,7 +862,7 @@ - (id)objectForKeyedSubscript:(NSString *)key
                   [self lock];
                       [[NSFileManager defaultManager] removeItemAtPath:[fileURL path] error:&error];
                   [self unlock];
-                  PINDiskCacheError(error);
+                  PINDiskCacheException(exception);
```
maicki: In this case should we not also print the error besides the exception?

```
PINDiskCacheError(error);
PINDiskCacheException(exception);
```

justin1121: Probably could, error will most likely be nil most of the time. Removing the file should succeed in most cases.

Source/PINDiskCache.m
```diff
@@ -16,6 +16,10 @@
 [[NSString stringWithUTF8String:__FILE__] lastPathComponent], \
 __LINE__, [error localizedDescription]); }
 
+#define PINDiskCacheException(exception) if (exception) { NSLog(@"%@ (%d) ERROR: %@", \
+[[NSString stringWithUTF8String:__FILE__] lastPathComponent], \
+__LINE__, [exception reason]); }
```
garrettmoon: I wonder if this should, instead of logging, `NSAssert(NO, [exception reason]);`? That way it will crash in debug but not in production (by default)? What do you think?

Source/PINDiskCache.m
garrettmoon: I agree with @maicki we should leave the PINDiskCacheError. We should probably move the PINDiskCacheException to the top of this scope as well?

Source/PINDiskCache.m
justin1121: Yeah I'm into that. Its a pretty serious error. I'll make the changes.

Source/PINDiskCache.m
justin1121: If we make the PINDiskCacheException do an assert instead, won't it crash before its able to remove the file and print the error? Or perhaps the file shouldn't be removed if it crashes in debug? Feel like we want to remove it in every case though.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
justin1121: Yeah that sounds good to me. I've actually already made the changes for #3. See below.

PR #198, opened 2017-09-20
[PINDiskCache] Respect small byteLimit settings by checking object size in setObject:

appleguy:
This serves as a mechanism for some clients, like users of the Texture framework, to
effectively disable the PINDiskCache by setting a byteLimit of 1.

Before this change, the cache would transiently exceed the byte limit by writing
the file anyway, and then immediately deleting it.

This change will not affect most users of PINDiskCache, but is important for this
specific use case.

### DIFF
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -838,7 +838,7 @@ static NSURL *_sharedTrashURL;
 {
     NSDate *now = [[NSDate alloc] init];
     
-    if (!key)
+    if (!key || (_dates.count == 0 && _sizes.count == 0))
         return nil;
     
     id <NSCoding> object = nil;
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -936,8 +936,23 @@ static NSURL *_sharedTrashURL;
       NSDataWritingOptions writeOptions = NSDataWritingAtomic;
     #endif
   
-    NSURL *fileURL = [self encodedFileURLForKey:key];
-    
+    // Remain unlocked here so that we're not locked while serializing.
+    NSData *data = _serializer(object, key);
+    NSURL *fileURL = nil;
+
+    NSUInteger byteLimit = self.byteLimit;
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
```diff
@@ -838,7 +838,7 @@ - (id)objectForKeyedSubscript:(NSString *)key
 {
     NSDate *now = [[NSDate alloc] init];
     
-    if (!key)
+    if (!key || (_dates.count == 0 && _sizes.count == 0))
```
garrettmoon: You need a lock here both because this is not thread-safe (at least not formally) and because the disk cache may not have been initialized yet.

Tests/PINCacheTests.m
```diff
@@ -877,7 +877,7 @@ - (void)testAsyncDiskInitialization
     
     testCache = [[PINDiskCache alloc] initWithName:cacheName];
     //This should not return until *after* disk cache directory has been created
-    [testCache objectForKey:@"some bogus key"];
```
garrettmoon: This actually shouldn't be changed, it failed because the optimization above, without the lock, isn't valid. I think it would cause problems for systems requesting objects early in their lifecycle which might be critical to pull from the cache.

appleguy:
@garrettmoon thanks for the feedback, I think I understand now! I'm happy to change this further if you see other issues.

PR #203, opened 2017-10-02
Significantly improves startup performance by asynchronously building…

garrettmoon:
… known state on startup.

### DIFF
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -458,6 +458,8 @@ static NSURL *_sharedTrashURL;
             _metadata[key].size = fileSize;
             byteCount += [fileSize unsignedIntegerValue];
         }
+        
+        [_knownKeys addObject:key];
     }
     
     if (byteCount > 0)
@@ -1013,6 +1015,8 @@ static NSURL *_sharedTrashURL;
             if (date) {
                 self->_metadata[key].date = date;
             }
+            
+            [_knownKeys addObject:key];
             
             if (self->_byteLimit > 0 && self->_byteCount > self->_byteLimit)
                 [self trimToSizeByDateAsync:self->_byteLimit completion:nil];
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -46,8 +46,6 @@ static PINOperationDataCoalescingBlock PINDiskTrimmingDateCoalescingBlock = ^id(
 @end
 
 @interface PINDiskCache () {
-    NSConditionLock *_instanceLock;
-    
     PINDiskCacheSerializerBlock _serializer;
     PINDiskCacheDeserializerBlock _deserializer;
     
@@ -55,11 +53,13 @@ static PINOperationDataCoalescingBlock PINDiskTrimmingDateCoalescingBlock = ^id(
     PINDiskCacheKeyDecoderBlock _keyDecoder;
 }
 
+@property (assign, nonatomic) pthread_mutex_t mutex;
 @property (copy, nonatomic) NSString *name;
 @property (assign) NSUInteger byteCount;
 @property (strong, nonatomic) NSURL *cacheURL;
 @property (strong, nonatomic) PINOperationQueue *operationQueue;
 @property (strong, nonatomic) NSMutableDictionary <NSString *, PINDiskCacheMetadata *> *metadata;
+@property (assign, nonatomic) BOOL diskWritable;
 @property (assign, nonatomic) BOOL diskStateKnown;
 @end
 
@@ -82,6 +82,12 @@ static NSURL *_sharedTrashURL;
 #endif
 
 #pragma mark - Initialization -
+
+- (void)dealloc
+{
+    __unused int result = pthread_mutex_destroy(&_mutex);
+    NSCAssert(result == 0, @"Failed to destroy lock in PINMemoryCache %p. Code: %d", (void *)self, result);
+}
 
 - (instancetype)init
 {
@@ -143,10 +149,12 @@ static NSURL *_sharedTrashURL;
               @"PINDiskCache must be initialized with a encoder AND decoder.");
     
     if (self = [super init]) {
+        __unused int result = pthread_mutex_init(&_mutex, NULL);
+        NSAssert(result == 0, @"Failed to init lock in PINMemoryCache %@. Code: %d", self, result);
+        
         _name = [name copy];
         _prefix = [prefix copy];
         _operationQueue = operationQueue;
-        _instanceLock = [[NSConditionLock alloc] initWithCondition:PINDiskCacheConditionNotReady];
         _willAddObjectBlock = nil;
         _willRemoveObjectBlock = nil;
         _willRemoveAllObjectsBlock = nil;
@@ -198,11 +206,10 @@ static NSURL *_sharedTrashURL;
 
         //we don't want to do anything without setting up the disk cache, but we also don't want to block init, it can take a while to initialize. This must *not* be done on _operationQueue because other operations added may hold the lock and fill up the queue.
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-            //should always be able to aquire the lock unless the below code is running.
-            [_instanceLock lockWhenCondition:PINDiskCacheConditionNotReady];
+            [self lock];
                 [self _locked_createCacheDirectory];
-                [self _locked_initializeDiskProperties];
-            [_instanceLock unlockWithCondition:PINDiskCacheConditionReady];
+            [self unlock];
+            [self initializeDiskProperties];
         });
     }
     return self;
@@ -415,8 +422,10 @@ static NSURL *_sharedTrashURL;
 
 - (BOOL)_locked_createCacheDirectory
 {
-    if ([[NSFileManager defaultManager] fileExistsAtPath:[_cacheURL path]])
+    if ([[NSFileManager defaultManager] fileExistsAtPath:[_cacheURL path]]) {
+        _diskWritable = YES;
         return NO;
+    }
     
     NSError *error = nil;
     BOOL success = [[NSFileManager defaultManager] createDirectoryAtURL:_cacheURL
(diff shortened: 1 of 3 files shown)


pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

appleguy:
Excited for this PR :). Let me know when it's ready for a full review — I'll try to look in depth within a day or two.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
```diff
@@ -424,45 +433,61 @@ - (BOOL)_locked_createCacheDirectory
...
         
         error = nil;
-        NSDictionary *dictionary = [fileURL resourceValuesForKeys:keys error:&error];
-        PINDiskCacheError(error);
         
-        NSDate *date = [dictionary objectForKey:NSURLContentModificationDateKey];
-        if (date && key)
-            [_dates setObject:date forKey:key];
+        // Do not continue to hold the lock while processing files.
+        [self lock];
```
appleguy: The comment doesn't seem to match the code - should this be re-locking or unlocking?

Suggestion that would avoid the many lock / unlock cycles:
- Build up a separate NSMutableDictionary instance (easier if merged into one, but can be done with multiple)
- No locking within the loop
- At the end of the loop, lock and merge into the new instances any keys and values that have been set on the _sizes / _dates / _knownKeys while loading from disk has been going on.
- Overwrite the instance variables with the new collections.

This would be the only way to avoid contention in the common case of a busy cache (e.g. many images loading at startup) while loading a large disk cache (e.g. a 200MB+ cache of 10-100KB images).

Source/PINDiskCache.m
```diff
@@ -424,45 +433,61 @@ - (BOOL)_locked_createCacheDirectory
...
     }
     
-    if (byteCount > 0)
-        _byteCount = byteCount;
+    [self lock];
+        if (byteCount > 0)
+            _byteCount = byteCount;
+    
+        if (self->_byteLimit > 0 && self->_byteCount > self->_byteLimit)
+            [self trimToSizeByDateAsync:self->_byteLimit completion:nil];
```
appleguy: It looks like we may need a way to avoid scheduling a large number of trim operations while this is processing. It probably makes sense to have a _trimPending / scheduled variable, so that a series of setObject: calls during this time doesn't schedule a lot of them.

Since the operation queue is concurrent, it looks like there may be an issue with multiple trim operations starting in parallel too — this could be addressed by using a barrier block on the underlying queue, which GCD will ensure is run serially.

Source/PINDiskCache.m
```diff
@@ -974,7 +1005,7 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key fileURL:(NSURL **
         return;
     }
 
-    [self lock];
+    [self lockUntilWritable];
```
appleguy: Is it necessary to block until the setup is done? If this happens, then won't the common path that triggers cache init (e.g. an image download starting from PINRemoteImage) will still block displaying the image until after the cache is fully read in?

I may be missing something, like perhaps PINRemoteImage performs the setObject: asynchronously and we won't block.

Source/PINDiskCache.m
```diff
@@ -854,17 +880,22 @@ - (id)objectForKeyedSubscript:(NSString *)key
...
         return nil;
     
     id <NSCoding> object = nil;
     NSURL *fileURL = [self encodedFileURLForKey:key];
     
     [self lock];
+        if (self->_ttlCache) {
+            // We actually need to know the entire disk state if we're a TTL cache.
+            [self unlock];
+            [self lockUntilDiskStateKnown];
```
appleguy: Why is that? Can't we just return nil for any objectForKey: before initialization is done, including for a TTL cache?

I think I recall there being a default TTL of 30 days. We should consider turning this off by default if it incurs a meaningful performance penalty, which blocking on cache init would be a pretty big cost if the developer hasn't specifically indicated they want TTL behavior.

Source/PINDiskCache.m
```diff
@@ -1389,20 +1421,52 @@ - (void)setWritingProtectionOption:(NSDataWritingOptions)writingProtectionOption
...
+{
+    __unused int result = pthread_mutex_lock(&_mutex);
+    NSAssert(result == 0, @"Failed to lock PINDiskCache %@. Code: %d", self, result);
+    
+    // spinlock if the disk isn't writable
+    while (_diskWritable == NO) {
+        [self unlock];
+        usleep(100);
+        
+        __unused int result = pthread_mutex_lock(&_mutex);
```
appleguy: Why not call [self lock] here and in the method below?

Source/PINDiskCache.m
```diff
@@ -1389,20 +1421,52 @@ - (void)setWritingProtectionOption:(NSDataWritingOptions)writingProtectionOption
...
 
+- (void)lockUntilWritable
+{
+    __unused int result = pthread_mutex_lock(&_mutex);
+    NSAssert(result == 0, @"Failed to lock PINDiskCache %@. Code: %d", self, result);
+    
+    // spinlock if the disk isn't writable
+    while (_diskWritable == NO) {
+        [self unlock];
+        usleep(100);
```
appleguy: Using this technique could cause some other problems, like a lot of threads to be spawned. We should be able to rely on the lock to be released in order to let the threads proceed.

How about a separate read-write mutex, just for cache init? At initialization of the class, the writer lock would be acquired by the code doing the read from disk. Any code calling -lockUntilWritable would first acquire the read lock (which could be grabbed by many threads at the same time, as long as the write lock isn't held). Secondly it would acquire the main [self lock].

This should allow all of them to wait on the initialization, without busying the main cache lock for operations that don't depend on _diskWritable.

As an alternative, I think there is a pthread version of the condition lock or a semaphore which can be signalled.

Source/PINDiskCache.m
garrettmoon: @appleguy the tim's are actually coalesced thanks to @nguyenhuy's improvement in PINOperationQueue! We could also skip trimming until we have a known state.

Source/PINDiskCache.m
garrettmoon: TTL is off by default. It was a feature added by a community member which *enforces* ttl (I.e. it won't return an object even if it has it if it exceeds the TTL). There are separate time limits are probably the default you're think of but they only relate to trimming, not a guarantee something won't be returned. To be frank I'm a bit regretful we merged in this feature, it's not really in the spirit of the framework and adds a lot of complexity.

Source/PINDiskCache.m
garrettmoon: lockUntilWritable only waits until the cache directory is confirmed to exist or is created if it doesn't. lockUntilKnownState (I think that's what I called it) guarantees everything is created, iterated and populated.

Source/PINDiskCache.m
garrettmoon: Good point.

Source/PINDiskCache.m
garrettmoon: Interesting, I'll play around with this.

Source/PINDiskCache.m
garrettmoon: Ah, the comment is supposed to mean, do not hold the lock the entire time while processing files.

The real issue around this locking is not the in memory state of ivars but disk access. PINDiskCache manages access to the disk solely through the lock. Without it, we can't prevent another operation from overwriting our file while we're trying to get its attributes.

In the future future, I really want to investigate using NSFileCoordinators instead of locking to manage disk access.

appleguy: Very interesting! I wonder if we should use a separate lock for disk access and for protecting the shared state? There are probably a bunch of methods that don't need to block on file access...

I think this explains why I've seen more lock contention in this layer than I would expect from just protecting the shared state. The good news is that, unlike some hierarchical objects like nodes where multiple locks create deadlocks extremely easily, this class might be ideally structured for a clean multi-lock strategy.

Source/PINDiskCache.m
garrettmoon: I think the pthread condition might be the best option. I'm *guessing* performance isn't any better for read / write locks and it's far less indicative of what we're doing. Need to actually figure out how to do pthread conditions though, not a lot of great docs…

Source/PINDiskCache.m
```diff
@@ -415,55 +427,78 @@ + (void)emptyTrash
...
-        return NO;
+    BOOL created = NO;
+    if ([[NSFileManager defaultManager] fileExistsAtPath:[_cacheURL path]] == NO) {
+        NSError *error = nil;
+        BOOL success = [[NSFileManager defaultManager] createDirectoryAtURL:_cacheURL
+                                                withIntermediateDirectories:YES
+                                                                 attributes:nil
+                                                                      error:&error];
+        PINDiskCacheError(error);
+        created = success;
```
Adlai-Holler: Nit: Let's assign directly to `created` and remove `success`

Source/PINDiskCache.m
```diff
@@ -1397,20 +1439,42 @@ - (void)setWritingProtectionOption:(NSDataWritingOptions)writingProtectionOption
...
+
+- (void)lockUntilDiskStateKnown
+{
+    [self lock];
+    
+    // spinlock if the disk state isn't known
+    if (_diskStateKnown == NO) {
+        pthread_cond_wait(&_diskStateKnownCondition, &_mutex);
+    }
+}
```
Adlai-Holler: Two alternate names might be `lockForWriting` and `lockAndWaitForKnownDiskState` or something. `lockUntil` isn't quite on the money.

Source/PINDiskCache.m
```diff
@@ -415,55 +427,78 @@ + (void)emptyTrash
...
-                                                   includingPropertiesForKeys:keys
-                                                                      options:NSDirectoryEnumerationSkipsHiddenFiles
-                                                                        error:&error];
+    
+    [self lock];
+        NSArray *files = [[NSFileManager defaultManager] contentsOfDirectoryAtURL:_cacheURL
+                                                       includingPropertiesForKeys:keys
+                                                                          options:NSDirectoryEnumerationSkipsHiddenFiles
+                                                                            error:&error];
+    [self unlock];
```
Adlai-Holler: What's the motivation behind this lock? `cacheURL` is immutable and I think it's safe to use NSFileManager from multiple threads.

Source/PINDiskCache.m
```diff
@@ -859,17 +895,22 @@ - (id)objectForKeyedSubscript:(NSString *)key
     NSDate *now = [[NSDate alloc] init];
```
Adlai-Holler: Unrelated to this diff, but I just noticed we can save some work by moving this date creation down.

Source/PINDiskCache.m
```diff
@@ -830,7 +866,7 @@ - (void)enumerateObjectsWithBlockAsync:(PINDiskCacheFileURLEnumerationBlock)bloc
 - (void)synchronouslyLockFileAccessWhileExecutingBlock:(PINCacheBlock)block
 {
     if (block) {
-        [self lock];
+        [self lockUntilWritable];
             block(self);
         [self unlock];
     }
```
Adlai-Holler: Should we also modify `containsObjectForKey:`?

Source/PINDiskCache.m
```diff
@@ -83,6 +85,12 @@ @implementation PINDiskCache
 
 #pragma mark - Initialization -
 
+- (void)dealloc
+{
+    __unused int result = pthread_mutex_destroy(&_mutex);
+    NSCAssert(result == 0, @"Failed to destroy lock in PINMemoryCache %p. Code: %d", (void *)self, result);
```
Adlai-Holler: Add `pthread_cond_destroy` here

Source/PINDiskCache.m
garrettmoon: I like those better.

Source/PINDiskCache.m
garrettmoon: I knew there had to be something like that (but couldn't find docs)

Source/PINDiskCache.m
garrettmoon: Good catch!

Source/PINDiskCache.m
garrettmoon: You're right that it's not strictly necessary and I think we should likely audit this behavior to further improve performance. However in it's current state, all access to the filesystem is done with the lock held and I want to keep that behavior intact until we decide on a cohesive replacement strategy.

Issue #31, opened 2015-09-26
Ability to purge expired items


kerrmarin:
Hi there,

It would be great to be able to purge items that have expired from the cache. For example, I have a service that returns images with a cache-control header. I would like to have the ability to, when inserting into the cache, specify a date to mark them as expired and then have the ability to remove only the expired items from the cache.

Thanks

garrettmoon:
Would be happy to accept a pull request that implemented this without a performance cost for use that didn't require this functionality.

ryanbrodie:
I second this, purging at the object level rather than the entire cache would be useful in many use cases.

garrettmoon:
@kerrmarin TTL has been implemented as an option. Does this address this issue?

PR #209, opened 2017-12-16
Add ability to set an object level TTL

mjlazar:
I think this partially addresses #31.

### DIFF
--- a/Source/PINCache.m
+++ b/Source/PINCache.m
@@ -233,6 +233,26 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     [group start];
 }
 
+- (void)removeExpiredObjectsAsync:(PINCacheBlock)block
+{
+    PINOperationGroup *group = [PINOperationGroup asyncOperationGroupWithQueue:_operationQueue];
+
+    [group addOperation:^{
+        [_memoryCache removeExpiredObjects];
+    }];
+    [group addOperation:^{
+        [_diskCache removeExpiredObjects];
+    }];
+
+    if (block) {
+        [group setCompletion:^{
+            block(self);
+        }];
+    }
+
+    [group start];
+}
+
 #pragma mark - Public Synchronous Accessors -
 
 - (NSUInteger)diskByteCount
@@ -328,6 +348,12 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     
     [_memoryCache trimToDate:date];
     [_diskCache trimToDate:date];
+}
+
+- (void)removeExpiredObjects
+{
+    [_memoryCache removeExpiredObjects];
+    [_diskCache removeExpiredObjects];
 }
 
 - (void)removeAllObjects
--- a/Source/PINCache.m
+++ b/Source/PINCache.m
@@ -45,6 +45,17 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
                   keyEncoder:(PINDiskCacheKeyEncoderBlock)keyEncoder
                   keyDecoder:(PINDiskCacheKeyDecoderBlock)keyDecoder
 {
+    return [self initWithName:name rootPath:rootPath serializer:serializer deserializer:deserializer keyEncoder:keyEncoder keyDecoder:keyDecoder ttlCache:NO];
+}
+
+- (instancetype)initWithName:(NSString *)name
+                    rootPath:(NSString *)rootPath
+                  serializer:(PINDiskCacheSerializerBlock)serializer
+                deserializer:(PINDiskCacheDeserializerBlock)deserializer
+                  keyEncoder:(PINDiskCacheKeyEncoderBlock)keyEncoder
+                  keyDecoder:(PINDiskCacheKeyDecoderBlock)keyDecoder
+                    ttlCache:(BOOL)ttlCache
+{
     if (!name)
         return nil;
     
@@ -60,7 +71,8 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
                                            deserializer:deserializer
                                              keyEncoder:nil
                                              keyDecoder:nil
-                                         operationQueue:_operationQueue];
+                                         operationQueue:_operationQueue
+                                               ttlCache:ttlCache];
         _memoryCache = [[PINMemoryCache alloc] initWithOperationQueue:_operationQueue];
     }
     return self;
--- a/Source/PINCache.m
+++ b/Source/PINCache.m
@@ -134,7 +134,17 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     [self setObjectAsync:object forKey:key withCost:0 completion:block];
 }
 
+- (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit completion:(PINCacheObjectBlock)block
+{
+    [self setObjectAsync:object forKey:key withCost:0 ageLimit:ageLimit completion:block];
+}
+
 - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withCost:(NSUInteger)cost completion:(PINCacheObjectBlock)block
+{
+    [self setObjectAsync:object forKey:key withCost:cost ageLimit:0.0 completion:block];
+}
+
+- (void)setObjectAsync:(nonnull id)object forKey:(nonnull NSString *)key withCost:(NSUInteger)cost ageLimit:(NSTimeInterval)ageLimit completion:(nullable PINCacheObjectBlock)block
 {
     if (!key || !object)
         return;
@@ -142,10 +152,10 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     PINOperationGroup *group = [PINOperationGroup asyncOperationGroupWithQueue:_operationQueue];
     
     [group addOperation:^{
-        [_memoryCache setObject:object forKey:key withCost:cost];
+        [_memoryCache setObject:object forKey:key withCost:cost ageLimit:ageLimit];
     }];
     [group addOperation:^{
-        [_diskCache setObject:object forKey:key];
+        [_diskCache setObject:object forKey:key withAgeLimit:ageLimit];
     }];
   
     if (block) {
@@ -269,13 +279,23 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     [self setObject:object forKey:key withCost:0];
 }
 
+- (void)setObject:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit
+{
+    [self setObject:object forKey:key withCost:0 ageLimit:ageLimit];
+}
+
 - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key withCost:(NSUInteger)cost
+{
+    [self setObject:object forKey:key withCost:cost ageLimit:0.0];
+}
+
+- (void)setObject:(nullable id)object forKey:(nonnull NSString *)key withCost:(NSUInteger)cost ageLimit:(NSTimeInterval)ageLimit
 {
     if (!key || !object)
         return;
     
-    [_memoryCache setObject:object forKey:key withCost:cost];
-    [_diskCache setObject:object forKey:key];
+    [_memoryCache setObject:object forKey:key withCost:cost ageLimit:ageLimit];
+    [_diskCache setObject:object forKey:key withAgeLimit:ageLimit];
 }
 
 - (nullable id)objectForKeyedSubscript:(NSString *)key
--- a/Source/PINCache.m
+++ b/Source/PINCache.m
@@ -238,10 +238,10 @@ static NSString * const PINCacheSharedName = @"PINCacheShared";
     PINOperationGroup *group = [PINOperationGroup asyncOperationGroupWithQueue:_operationQueue];
 
     [group addOperation:^{
-        [_memoryCache removeExpiredObjects];
+        [self->_memoryCache removeExpiredObjects];
     }];
     [group addOperation:^{
-        [_diskCache removeExpiredObjects];
+        [self->_diskCache removeExpiredObjects];
     }];
 
     if (block) {
--- a/Source/PINDiskCache.h
+++ b/Source/PINDiskCache.h
@@ -362,7 +362,8 @@ PIN_SUBCLASSING_RESTRICTED
 
  @param object An object to store in the cache.
  @param key A key to associate with the object. This string will be copied.
- @param ageLimit The age limit (in seconds) to associate with the object.
+ @param ageLimit The age limit (in seconds) to associate with the object. An age limit <= 0 means there is no object-level age limit and the cache-level TTL
+                 will be used for this object.
  @param block A block to be executed serially after the object has been stored, or nil.
  */
 - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit completion:(nullable PINDiskCacheObjectBlock)block;
@@ -389,7 +390,8 @@ PIN_SUBCLASSING_RESTRICTED
  @param object An object to store in the cache.
  @param key A key to associate with the object. This string will be copied.
  @param cost An amount to add to the <memoryCache.totalCost>.
- @param ageLimit The age limit (in seconds) to associate with the object.
+ @param ageLimit The age limit (in seconds) to associate with the object. An age limit <= 0 means there is no object-level age limit and the cache-level TTL will be used for
+                 this object.
  @param block A block to be executed concurrently after the object has been stored, or nil.
  */
 - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withCost:(NSUInteger)cost ageLimit:(NSTimeInterval)ageLimit completion:(nullable PINCacheObjectBlock)block;
@@ -489,7 +491,8 @@ PIN_SUBCLASSING_RESTRICTED
  @see setObjectAsync:forKey:completion:
  @param object An object to store in the cache.
  @param key A key to associate with the object. This string will be copied.
- @param ageLimit The age limit (in seconds) to associate with the object.
+ @param ageLimit The age limit (in seconds) to associate with the object. An age limit <= 0 means there is
+                 no object-level age limit and the cache-level TTL will be used for this object.
  */
 - (void)setObject:(nullable id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit;
 
--- a/Source/PINDiskCache.h
+++ b/Source/PINDiskCache.h
@@ -13,7 +13,13 @@ NS_ASSUME_NONNULL_BEGIN
 @class PINDiskCache;
 @class PINOperationQueue;
 
+extern NSString * const PINDiskCacheErrorDomain;
+extern NSErrorUserInfoKey const PINDiskCacheErrorWriteFailureCodeKey;
 extern NSString * const PINDiskCachePrefix;
+
+typedef NS_ENUM(NSInteger, PINDiskCacheError) {
+  PINDiskCacheErrorWriteFailure = -1000,
+};
 
 /**
  A callback block which provides the cache, key and object as arguments
@@ -351,6 +357,17 @@ PIN_SUBCLASSING_RESTRICTED
 - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key completion:(nullable PINDiskCacheObjectBlock)block;
 
 /**
+ Stores an object in the cache for the specified key and age limit. This method returns immediately and executes the
+ passed block as soon as the object has been stored.
+
+ @param object An object to store in the cache.
+ @param key A key to associate with the object. This string will be copied.
+ @param ageLimit The age limit (in seconds) to associate with the object.
+ @param block A block to be executed serially after the object has been stored, or nil.
+ */
+- (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit completion:(nullable PINDiskCacheObjectBlock)block;
+
+/**
  Stores an object in the cache for the specified key and the specified memory cost. If the cost causes the total
  to go over the <memoryCache.costLimit> the cache is trimmed (oldest objects first). This method returns immediately
  and executes the passed block after the object has been stored, potentially in parallel with other blocks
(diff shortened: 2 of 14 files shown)


Source/PINDiskCache.m
```diff
@@ -881,8 +939,9 @@ - (id)objectForKeyedSubscript:(NSString *)key
             [self unlock];
             [self lockAndWaitForKnownState];
         }
-    
-        if (!self->_ttlCache || self->_ageLimit <= 0 || fabs([_metadata[key].date timeIntervalSinceDate:now]) < self->_ageLimit) {
+
+        NSTimeInterval ageLimit = _metadata[key].hasAgeLimit ? _metadata[key].ageLimit : self->_ageLimit;
```
garrettmoon: Same question as above.

Source/PINDiskCache.m
```diff
@@ -881,8 +939,9 @@ - (id)objectForKeyedSubscript:(NSString *)key
...
+        if (!self->_ttlCache || ageLimit <= 0 || fabs([_metadata[key].date timeIntervalSinceDate:now]) < ageLimit) {
```
garrettmoon: Is this different behavior than above? I.e. it behaves like a ttlCache if there's an age limit *or* it's marked as a ttlCache here but not above?

Source/PINDiskCache.m
```diff
@@ -1025,7 +1094,9 @@ - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key fileURL:(NSURL **
             if (date) {
                 self->_metadata[key].date = date;
             }
-            
+            if (ageLimit > 0.0) {
+                [self asynchronouslySetAgeLimit:ageLimit forURL:fileURL];
```
garrettmoon: Not sure the answer to this: will there be implications to this being asynchronous? Specifically that _metadata[key].ageLimit won't be set immediately and checks against it could happen before it is?

I'm *guessing* the answer is no, but maybe there could be a comment here suggesting it's been thought through and why it's safe?

Source/PINDiskCache.m
```diff
@@ -1453,4 +1525,10 @@ - (void)enumerateObjectsWithBlock:(PINDiskCacheFileURLBlock)block completionBloc
 @end
 
 @implementation PINDiskCacheMetadata
+
+- (void)setAgeLimit:(NSTimeInterval)ageLimit {
+    _ageLimit = ageLimit;
+    _hasAgeLimit = YES;
```
garrettmoon: Would it be better to just make a getter for `hasAgeLimit` which checks `_ageLimit > 0`?

Source/PINDiskCache.m
```diff
@@ -1170,6 +1205,32 @@ - (void)trimToSizeByDate:(NSUInteger)trimByteCount
...
+        NSMutableArray<NSString *> *expiredObjectKeys = [NSMutableArray array];
+        [_metadata enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull key, PINDiskCacheMetadata * _Nonnull obj, BOOL * _Nonnull stop) {
+            NSTimeInterval ageLimit = obj.ageLimit > 0.0 ? obj.ageLimit : self->_ageLimit;
+            NSDate *expirationDate = [obj.date dateByAddingTimeInterval:ageLimit];
+            if ([expirationDate compare:now] == NSOrderedAscending) { // Expiration date has passed
+                [expiredObjectKeys addObject:key];
+            }
+        }];
+
+        for (NSString *key in expiredObjectKeys) {
```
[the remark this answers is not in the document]
garrettmoon: https://github.com/pinterest/PINCache/pull/212

Source/PINMemoryCache.m
```diff
@@ -196,6 +200,27 @@ - (void)trimMemoryToDate:(NSDate *)trimDate
     }
 }
 
+- (void)removeExpiredObjects
```
garrettmoon: Sorry this comment is in the wrong place (you're using dates[key] below which gets updated on access in objectForKey), but, should objectForKey: not set the access date if an ageLimit is set?

garrettmoon:
Thanks for writing the tests!

Source/PINCache.m
```diff
@@ -134,18 +134,28 @@ - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key completion:(
...
 }
 
+- (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit completion:(PINCacheObjectBlock)block
+{
+    [self setObjectAsync:object forKey:key withCost:0 ageLimit:ageLimit completion:block];
+}
+
 - (void)setObjectAsync:(id <NSCoding>)object forKey:(NSString *)key withCost:(NSUInteger)cost completion:(PINCacheObjectBlock)block
+{
+    [self setObjectAsync:object forKey:key withCost:cost ageLimit:0.0 completion:block];
```
garrettmoon: Once this lands, I wonder if we could reduce the complexity of everything by changing this line to:
`[self setObjectAsync:object forKey:key withCost:cost ageLimit:self.isTTLCache ? self.ageLimit : 0.0 completion:block]`

Then we could get rid of any other logic that deals with isTTLCache?

Source/PINDiskCache.m
```diff
@@ -485,6 +491,11 @@ - (void)initializeDiskProperties
                 _metadata[key].size = fileSize;
                 byteCount += [fileSize unsignedIntegerValue];
             }
+
+            NSTimeInterval ageLimit;
```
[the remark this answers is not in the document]
mjlazar: Added code to handle the failure.

So, I've spent some time doing some analysis and trying out different performance tweaks (which is why I haven't responded to your comments yet). Since PINDiskCache.ttlCache is not known at disk initialization, we cannot skip over the code which reads the extended file system attributes from disk.

There is a performance hit that will be incurred on cache initialization and is a function of the number of files in the cache. I tested adding 1000 objects into the disk cache and then compared initialization times between the current code and the code within this pull request. On average it took ~24ms to initialize the cache with the current code. On average it took ~35ms to initialize after the code changes, which is ~11ms/44% slower.

The slowdown might not be worth it for users that do not take advantage of the object-level age limits, so I have a proposal for mitigation:
- Make PINDiskCache.ttlCache (and PINDiskCache.ageLimit?) a readonly property (which will obviously break people who are currently using it).
- Add a ttlCache parameter to PINDiskCache's designated initializer.
- Skip over reading the extended file system attributes if ttlCache == NO. 

This change will basically have zero effect on users who are not using ttlCache.

Source/PINDiskCache.m
```diff
@@ -455,6 +455,7 @@ - (BOOL)_locked_createCacheDirectory
 - (void)initializeDiskProperties
 {
     NSUInteger byteCount = 0;
+    BOOL hasAtLeastOneAgeLimit = NO;
```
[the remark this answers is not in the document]
mjlazar: The value of isTTLCache is unset at this point. See my other comment about adding this as a param to the designated initializer.

Source/PINDiskCache.m
```diff
@@ -673,7 +697,7 @@ - (void)trimDiskToDate:(NSDate *)trimDate
         
         for (NSString *key in keysSortedByDate) { // oldest files first
             NSDate *accessDate = _metadata[key].date;
-            if (!accessDate)
+            if (!accessDate || _metadata[key].ageLimit > 0.0)
```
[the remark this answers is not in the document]
mjlazar: Yeah, I agree it is sort of confusing. My thoughts here are that the `trimDate` that is passed in here is based on the cache-level age limit and we don't want to remove objects that have overridden the age limit.

I don't know what the correct thing to do here would be. -trimToAgeLimitRecursively calling -removeExpiredObjectsAsync would certainly be an option, but I wanted to avoid over-calling it since it has to iterate over all of the objects in the cache to determine if they are expired or not. Although, I don't think there is much harm in letting these expired objects lay dormant on the filesystem until the next initialization; the getters/iterators will treat them as if they don't exist.

Source/PINDiskCache.m
```diff
@@ -845,8 +898,13 @@ - (BOOL)containsObjectForKey:(NSString *)key
 {
     [self lock];
         if (_metadata[key] != nil || _diskStateKnown == NO) {
+            BOOL objectExpired = NO;
+            if (self->_ttlCache && _metadata[key].date != nil) {
```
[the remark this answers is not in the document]
mjlazar: Yes, ageLimit only applies if ttlCache is true. I added an assertion as you suggested.

Source/PINDiskCache.m
mjlazar: I think this is basically saying:
- If ttlCache == NO, return the object.
- If ttlCache == YES and ageLimit is not set, return the file.
- If ttlCache == YES, ageLimit is set and object is not expired return the object.

Source/PINMemoryCache.m
mjlazar: Hmm, good point. The disk cache checks if object && !self->_ttlCache before setting the access date. This logic currently exists; I wonder if that's a bug. Shall I update the memory cache to do the same check?

Source/PINCache.m
mjlazar: I think that could certainly be done, but it would have to be in conjunction with some changes to the trim-to-date methods. Since this would essentially set object-level age limits all of the time, we could combine the current trimming logic with the new -removeExpiredObjects logic I'm proposing in this PR.

Source/PINDiskCache.m
```diff
@@ -553,15 +566,26 @@ - (void)asynchronouslySetAgeLimit:(NSTimeInterval)ageLimit forURL:(NSURL *)fileU
 
 - (BOOL)_locked_setAgeLimit:(NSTimeInterval)ageLimit forURL:(NSURL *)fileURL
 {
-    if (ageLimit <= 0.0 || !fileURL) {
+    if (!fileURL) {
         return NO;
```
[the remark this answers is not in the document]
mjlazar: I wasn't sure what the implications of asserting here. -_locked_setFileModificationDate does a similar nil check/return early. WDTY?

Source/PINDiskCache.m
```diff
@@ -1069,8 +1069,6 @@ - (void)setObject:(id)object forKeyedSubscript:(NSString *)key
 
 - (void)setObject:(id <NSCoding>)object forKey:(NSString *)key withAgeLimit:(NSTimeInterval)ageLimit fileURL:(NSURL **)outFileURL
 {
-    NSAssert(ageLimit <= 0.0 || (ageLimit > 0.0 && _ttlCache), @"ttlCache must be set to YES if setting an object-level age limit.");
```
mjlazar: I had to remove this assertion because `_ttlCache` is set asynchronously. If you tried to call -setObject immediately after instantiating the cache, the assertion would throw.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/bc733b55f5dffd387390ffa861a57c3f/raw/af1764c2fe339f8bcc8ca47f344922f545c1b4fa/log.txt)

Source/PINDiskCache.m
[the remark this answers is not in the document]
mjlazar: Done.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/f5e78e9cfc9a62de67dca3fa0a6417cd/raw/9e5e170d3fd9434fe08950e91347e68c6fa5435a/log.txt)

garrettmoon:
@mjlazar sorry, failures in CI weren't yours, I'm looking into it and will re-run when I think I've solved it.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/d602ed3e975a0d38df5841b45e246749/raw/20a9638f8d372259b38a4a613c357a75bf8f8c71/log.txt)

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/110cb14c1cae37b6a3f877ab74334a1e/raw/8c85aa03eb1041de115ce26da8f9bf42fda28abc/log.txt)

garrettmoon:
@mjlazar ok, latest log appears to be accurate (though the warnings are new to the upgraded CI).

mjlazar:
Warnings should be fixed now.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

mjlazar:
@garrettmoon In one of my previous comments, I proposed passing the ttlCache switch to the initializer(s) and making the property readonly. I just pushed a commit that implements that. PTAL; this would be a breaking change to folks that are utilizing ttlCache.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/5e65d136c6c95d73116425808ac7f9bc/raw/cbd85c7acdf06ab8e2bfeee396fa735e453f915e/log.txt)

mjlazar:
@garrettmoon is there any chance the latest CI failure is due to a flaky test? I can't reproduce locally.

Source/PINDiskCache.m
mjlazar: My thoughts were if `ageLimit` were set to anything (including zero), it would override the cache's `ageLimit`. I realize this is actually confusing, so I removed `hasAgeLimit`; the object's `ageLimit` must now be greater than zero in order to override the cache's `ageLimit`.

Source/PINDiskCache.m
mjlazar: Obsolete. `hasAgeLimit` has been removed.

garrettmoon:
@mjlazar I wasn't able to repro either so I'll re-run. I've never seen that one fail due to flakiness though…

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
```diff
@@ -1116,6 +1254,30 @@ - (void)trimToSizeByDate:(NSUInteger)trimByteCount
...
+        [_metadata enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull key, PINDiskCacheMetadata * _Nonnull obj, BOOL * _Nonnull stop) {
+            NSTimeInterval ageLimit = obj.ageLimit > 0.0 ? obj.ageLimit : self->_ageLimit;
+            NSDate *expirationDate = [obj.date dateByAddingTimeInterval:ageLimit];
+            if ([expirationDate compare:now] == NSOrderedAscending) { // Expiration date has passed
+                [expiredObjectKeys addObject:key];
+            }
+        }];
+
+        [self unlock];
+            for (NSString *key in expiredObjectKeys) {
```
garrettmoon: Can this for-loop be moved outside the lock? Or the `[self lock]` below should be changed to `[self lockForWriting]`?

Source/PINMemoryCache.m
garrettmoon: I'm wondering which is actually the bug…

`trimDiskToSizeByDate` and `trimToCostLimitByDate` both use last access to trim to order which objects they trim first. If ttlCache is enabled or there are age limits, objects which exceed the age limit should be trimmed first and then objects should be trimmed by last access. Does that seem right? In which case the two methods will need to be updated and the diskCache will need to remove its check of `object && !self->_ttlCache`…

Source/PINDiskCache.m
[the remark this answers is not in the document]
garrettmoon: I think the assert got removed in a follow up commit?

Source/PINMemoryCache.m
```diff
@@ -193,6 +206,27 @@ - (void)trimMemoryToDate:(NSDate *)trimDate
...
+{
+    [self lock];
+        NSDictionary<NSString *, NSDate *> *dates = [_dates copy];
+        NSDictionary<NSString *, NSNumber *> *ageLimits = [_ageLimits copy];
+    [self unlock];
+
+    NSDate *now = [NSDate date];
+    for (NSString *key in ageLimits) {
+        NSDate *accessDate = dates[key];
+        NSTimeInterval ageLimit = [_ageLimits[key] doubleValue] ?: self->_ageLimit;
```
garrettmoon: Accessing these ivars is unsafe, gather `self->_ageLimit` as `globalAgeLimit` in the lock above and use `ageLimits` here?

Source/PINDiskCache.m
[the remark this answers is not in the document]
garrettmoon: Ok, that makes sense, I think if we address the issue I mention below regarding `trimDiskToSizeByDate` and `trimToCostLimitByDate`

Source/PINDiskCache.m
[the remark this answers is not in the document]
mjlazar: I removed the `hasAtLeastOneAgeLimit` var all together since the `isTTLCache` check is all we need.

Source/PINDiskCache.m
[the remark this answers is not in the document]
mjlazar: Done.

Source/PINMemoryCache.m
mjlazar: Made the updates you suggested. PTAL.

Source/PINDiskCache.h
```diff
@@ -280,6 +292,29 @@ PIN_SUBCLASSING_RESTRICTED
  */
 - (instancetype)initWithName:(nonnull NSString *)name rootPath:(nonnull NSString *)rootPath serializer:(nullable PINDiskCacheSerializerBlock)serializer deserializer:(nullable PINDiskCacheDeserializerBlock)deserializer operationQueue:(nonnull PINOperationQueue *)operationQueue __attribute__((deprecated));
 
+/**
+ The designated initializer allowing you to override default NSKeyedArchiver/NSKeyedUnarchiver serialization.
```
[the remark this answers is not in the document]
mjlazar: Done. Removed.

Source/PINDiskCache.m
mjlazar: Yes. Moved this outside the lock.

Source/PINMemoryCache.m
mjlazar: Done.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

Source/PINDiskCache.m
```diff
@@ -1116,6 +1257,26 @@ - (void)trimToSizeByDate:(NSUInteger)trimByteCount
...
 }
 
+- (void)removeExpiredObjects
+{
+    [self lockForWriting];
+        NSDate *now = [NSDate date];
+        NSMutableArray<NSString *> *expiredObjectKeys = [NSMutableArray array];
+        [_metadata enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull key, PINDiskCacheMetadata * _Nonnull obj, BOOL * _Nonnull stop) {
+            NSTimeInterval ageLimit = obj.ageLimit > 0.0 ? obj.ageLimit : self->_ageLimit;
+            NSDate *expirationDate = [obj.date dateByAddingTimeInterval:ageLimit];
```
garrettmoon: Ok, so I think there's a problem: now we're always setting date. It should probably be renamed lastModifiedDate and a new property createdDate should be added? Because objects shouldn't have their expiration lengthened when they're accessed, but they should live longer, in relation to other objects, than other objects which haven't expired but were accessed later? This behavior will have to be added / updated in PINMemoryCache as well?

mjlazar: Good call. I modified the code using your suggestions. PTAL when you get a chance.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

garrettmoon:
This is looking great! One final request: would you be willing to update the tests to exercise our new creation / modification logic?

mjlazar:
@garrettmoon, I added some additional tests that exercise the new logic. Let me know what you think. Thanks!

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

pinterest-ios:
🚫 CI failed with [log](https://gist.githubusercontent.com/pinterest-ios/6db30203f6b75e6f7da497f28bcb2dc3/raw/80909c84e514e440c03ed5e75a0f97b153e00ab1/log.txt)

mjlazar:
Seems to be a race condition. I'll push a fix shortly.

mjlazar:
@garrettmoon, would you mind trying the tests again? Thanks.

pinterest-ios:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This is a big PR, please consider splitting it up to ease code review.</td>
    </tr>
  </tbody>
</table>

</p>

mjlazar:
Awesome! Thanks for reviewing.

PR #222, opened 2018-04-16
Fix mem cache TTL

jerelevine:
This fixes an issue where the memory cache was updating the last fetched date even when the cache is using TTL. The disk cache works this way and the memory cache should function in a comparable way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
-- [fix] Fix PINMemoryCache to check if TTL before updating last fetched date: [#222](https://github.com/pinterest/PINCache/pull/222) (jerelevine)
+- [fix] Fix PINMemoryCache to check TTL before updating last fetched date: [#222](https://github.com/pinterest/PINCache/pull/222) (jerelevine)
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## master
 
 * Add your own contributions to the next release on the line below this with your name.
+- [fix] Fix PINMemoryCache to check if TTL before updating last fetched date: [#222](https://github.com/pinterest/PINCache/pull/222) (jerelevine)
```
jparise: Should this be just "check TTL"? The "if" seems out of place.

jerelevine: Ok

jerelevine: Removed the unnecessary "if"

jerelevine:
How do we feel about this bug fix?

garrettmoon:
@jerelevine sorry for the slow response! It's a good bug fix for the current behavior!

However I'm not sure that behavior is actually correct, there's a lot of discussion here if you're interested in participating: https://github.com/pinterest/PINCache/pull/209

garrettmoon:
@jerelevine now that #209 has landed, is this PR necessary?

garrettmoon:
I'm going to close this out, lemme know if you want to re-open!

PR #266, opened 2020-03-20
allow experiments in PINCache

vovasty:
This PR introduces global configuration for `PINCache`. It can be used for experimental features.

### DIFF
--- a/Source/PINCacheConfiguration.h
+++ b/Source/PINCacheConfiguration.h
@@ -1,0 +1,29 @@
+//
+//  PINCacheConfiguration.h
+//  PINCache
+//
+//  Created by Vladimir Solomenchuk on 3/19/20.
+//  Copyright © 2020 Pinterest. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ * A bit mask of features.
+ */
+typedef NS_OPTIONS(NSUInteger, PINCacheExperimentalFeatures) {
+    PINCacheExperimentalFeaturesNone,
+    PINCacheExperimentalFeaturesAll = 0xFFFFFFFF
+};
+
+@interface PINCacheConfiguration: NSObject
+@end
+
+/**
+ * Implement this method in a category to make your
+ * configuration available to PINCache. It will be read
+ * only once and copied.
+ */
+@interface PINCacheConfiguration (UserProvided)
++ (PINCacheExperimentalFeatures)cacheConfiguration;
+@end
--- a/Source/PINCacheConfigurationInternal.h
+++ b/Source/PINCacheConfigurationInternal.h
@@ -1,0 +1,17 @@
+//
+//  PINCacheConfigurationInternal.h
+//  PINCache
+//
+//  Created by Vladimir Solomenchuk on 3/19/20.
+//  Copyright © 2020 Pinterest. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "PINCacheConfiguration.h"
+
+extern BOOL PINCacheHasActiveExperimentalFeature(PINCacheExperimentalFeatures option);
+
+@interface PINCacheConfigurationManager : NSObject
+// only for tests
++ (void) test_resetWithConfiguration:(PINCacheExperimentalFeatures) features;
+@end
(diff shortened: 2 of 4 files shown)


Source/PINCacheConfiguration.h
```diff
@@ -0,0 +1,28 @@
...
+//  Copyright © 2020 Pinterest. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+/**
+ * A bit mask of features.
+ */
+typedef NS_OPTIONS(NSUInteger, PINCacheExperimentalFeatures) {
+  PINCacheExperimentalFeatureAll = 0xFFFFFFFF
```
garrettmoon: Should we default to none instead of all?

Source/PINCacheConfigurationInternal.m
```diff
@@ -0,0 +1,63 @@
...
+
+#import "PINCacheConfigurationInternal.h"
+
+@interface PINCacheConfiguration(PINCacheConfiguration)
++ (PINCacheExperimentalFeatures) defaultConfiguration;
+@end
+
+@implementation PINCacheConfiguration
++ (PINCacheExperimentalFeatures) defaultConfiguration {
+    return PINCacheExperimentalFeatureAll;
```
garrettmoon: None instead of all?

Source/PINCacheConfigurationInternal.h
```diff
@@ -0,0 +1,17 @@
...
+//  PINCache
+//
+//  Created by Vladimir Solomenchuk on 3/19/20.
+//  Copyright © 2020 Pinterest. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "PINCacheConfiguration.h"
+
+extern BOOL PINCacheIsActiveExperimentalFeature(PINCacheExperimentalFeatures option);
```
jparise: I think `Has` might sound better than `Is` here?

Source/PINCacheConfigurationInternal.h
```diff
@@ -0,0 +1,17 @@
...
+
+@interface PINCacheConfigurationManager : NSObject
+// only for tests
++ (void) test_resetWithConfiguration:(PINCacheExperimentalFeatures) features;
```
garrettmoon: Are the tests ever called?

Source/PINCacheConfigurationInternal.m
```diff
@@ -0,0 +1,63 @@
...
+    if ([PINCacheConfiguration respondsToSelector:@selector(cacheConfiguration)]) {
+      _activatedExperiments = [PINCacheConfiguration cacheConfiguration];
+    } else {
+      _activatedExperiments = [PINCacheConfiguration defaultConfiguration];
+    }
+  }
+  return self;
+}
+
+- (BOOL)activateExperimentalFeature:(PINCacheExperimentalFeatures)requested
```
garrettmoon: Is this method activating the experiment? Or indicating whether it should be active?

Source/PINCacheConfigurationInternal.h
vovasty: not yet. It will be used in unit tests, when you want to test experimental functionality.

Source/PINCacheConfigurationInternal.m
vovasty: this method checks is an experiment is active. In code developers should use `PINCacheHasActiveExperimentalFeature` function. e.g.
```objc
if (PINCacheHasActiveExperimentalFeature(PINCacheHasActiveExperimentalFeatureDefaultDoNotRemoveAllObjectOnMemoryWarning)) {
   _removeAllObjectsOnMemoryWarning = NO;
} else {
   _removeAllObjectsOnMemoryWarning = YES;
}
```

vovasty:
@jparise we. can experiment with default values (that's what actually I want to do). Other options include alternative functionality.
Hooks and knobs work, but experiments allow you clearly define experimental areas and in large applications, that have a lot of caches, tunings to each cache might be complicated.

garrettmoon:
I actually agree with Jon that we may not want to configure PINCache globally. And that it's probably more valuable to tune the caches individually at this time and keep experimenting out of the framework. We only have 3 caches including the default at Pinterest, I don't think there's a concern for needing to tune lots of caches generally.

PR #269, opened 2020-04-13
Add Carthage for watchOS, fix macOS min deployment target version

### LINKED ISSUE #268: PINCache's Carthage seems does not support watchOS
Hi.

I'm the maintainer of [SDWebImage](https://github.com/SDWebImage/SDWebImage), and I'd like to provide a custom cache integration plugin of PINCache into SD's cache system.

The plugin is here: https://github.com/SDWebImage/SDWebImagePINPlugin

However, although the readme point the PINCache support watchOS, and the CocoaPods one support as well. For Carthage, it does not support watchOS. Because of the Xcode scheme does have any configuration for watchOS, even no any target. 

See log:

```
➜  SDWebImagePINPlugin git:(master) ✗ carthage update --no-use-binaries --platform watchos
*** Fetching SDWebImage
*** Fetching PINCache
*** Fetching PINOperation
*** Checking out SDWebImage at "5.7.2"
*** Checking out PINCache at "3.0.1-beta.8"
*** Checking out PINOperation at "1.1.2"
*** xcodebuild output can be found in /var/folders/y4/_98b424x3jv54_1tbfvswjpm0000gn/T/carthage-xcodebuild.lY4DVH.log
*** Skipped building PINOperation due to the error:
Dependency "PINOperation" has no shared framework schemes for any of the platforms: watchOS
```

dreampiggy:
Dependent https://github.com/pinterest/PINOperation/pull/31
This close #268 

CC: @garrettmoon

To make Carthage works (because you use ~> `1.1.0` dependency): 
You should firstly merge that PINOperation, release a new version.

Then merge this PR, release a new version.

Note: I also found that `Carthage/Checkouts` is not in Git Ignore files, does this have anything useful ? Why not ignore those files because when you run `carthage update`, it will be override and cause massive git diffs.


dreampiggy:
Local build success (I modify the Cartfile to point to my fork branch):

image

garrettmoon:
@dreampiggy Thanks for putting this up! I'm going to merge and put up another PR to update Carthage to use PINOperation 1.2.

With regards to the checkins, I'm not 1000% convinced it's the best course of action, but the intention is to make it so the repo can be built as is without fetching dependencies.

dreampiggy:
Finally...This get merged :)

PR #272, opened 2020-07-28
Support Catalyst

cgmaier:
- just checks a box : )


jparise:
@cgmaier Thanks! Is there anything we should add to our integration tests to verify Catalyst support?

cgmaier:
> @cgmaier Thanks! Is there anything we should add to our integration tests to verify Catalyst support?

not that I'm aware of - this should suffice : )

Issue #275, opened 2020-08-07
PINCache doesn't build in Xcode 12 Beta 4


mycroftcanner:
It generates a lot of semantic issues:

> Incompatible block pointer types sending 'PINMemoryCacheObjectBlock  _Nullable __strong' (aka 'void (^__strong)(PINMemoryCache * _Nonnull __strong, NSString * _Nonnull __strong, id  _Nullable __strong)') to parameter of type 'PINCacheObjectBlock _Nonnull' (aka 'void (^)(id<PINCaching>  _Nonnull __strong, NSString * _Nonnull __strong, id  _Nullable __strong)')

richieWave:
This

mycroftcanner:
It builds now after this [commit](https://github.com/pinterest/PINCache/pull/273/files/3267654d1d24fdb22bfb61aaa62b2d2e511554f0) but tests fail.

PINMemoryCache is also broken

garrettmoon:
@mycroftcanner Tests succeed for me locally, can you explain which tests fail and what you mean by PINMemoryCache is broken?

arangato:
I had a bunch of compilation errors in tests as well. All looking like the one @mycroftcanner initially reported. I've fixed it in [this PR](https://github.com/pinterest/PINCache/pull/276).

garrettmoon:
@mycroftcanner I just merged @arangato 's PR (thank you!), does this address what you were hitting?

Anatol1986:
The same issues also available on Xcode Version 12.0 beta 5

mycroftcanner:
@garrettmoon CocoaPods still look for old version and you specify the master branch then PINRemoteImage is out of sync and Texture won't build too.

bluemachine:
Issue still persists in Xcode 12 beta 5 with following pods
PINCache (3.0.1-beta.8)
PINOperation (1.1.2)
PINRemoteImage (3.0.0)

![Screenshot 2020-08-28 at 23 49 25](https://user-images.githubusercontent.com/880597/91617820-27751f80-e989-11ea-9440-ebed8e5d839d.png)

bluemachine:
Upgraded to Xcode 12 beta 6 and getting just this 1 issue. Hoping you could fix it soon. 

![Screenshot 2020-08-29 at 16 25 24](https://user-images.githubusercontent.com/880597/91639088-58e6fd00-ea14-11ea-8d38-e669cf1ece5c.png)

[...]

PR #281, opened 2020-09-08
Fix PINCaching compiling in Xcode 12.0b6 (#275)

sagesse-cn:
Reference: https://reviews.llvm.org/D66831#2019125

### DIFF
--- a/Source/PINCaching.h
+++ b/Source/PINCaching.h
@@ -17,18 +17,18 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  A callback block which provides only the cache as an argument
  */
-typedef void (^PINCacheBlock)(id<PINCaching> cache);
+typedef void (^PINCacheBlock)(__kindof id<PINCaching> cache);
 
 /**
  A callback block which provides the cache, key and object as arguments
  */
-typedef void (^PINCacheObjectBlock)(id<PINCaching> cache, NSString *key, id _Nullable object);
+typedef void (^PINCacheObjectBlock)(__kindof id<PINCaching> cache, NSString *key, id _Nullable object);
 
 /**
  A callback block used for enumeration which provides the cache, key and object as arguments plus a stop flag that
  may be flipped by the caller.
  */
-typedef void (^PINCacheObjectEnumerationBlock)(id<PINCaching> cache, NSString *key, id _Nullable object, BOOL *stop);
+typedef void (^PINCacheObjectEnumerationBlock)(__kindof id<PINCaching> cache, NSString *key, id _Nullable object, BOOL *stop);
 
 /**
  A callback block which provides a BOOL value as argument


garrettmoon:
@sagesse-cn I'm not seeing any errors compiling on Xcode 12 beta 6. Can you provide more info?

garrettmoon:
Specifically, if you could point out how I can reproduce it, that would be very helpful!

bluemachine:
When we install the pod PINRemoteImage - it installs with following pods

Installing PINCache (3.0.1-beta.8)
Installing PINOperation (1.1.2)
Installing PINRemoteImage (3.0.0)

To update as suggested - when we install PINCache it updates to latest version but installs an older version of PINRemoteImage and followings pods.

Installing FLAnimatedImage (1.0.12)
Installing PINCache 3.0.1 (was 3.0.1-beta.8)
Installing PINOperation 1.2 (was 1.1.2)
Installing PINRemoteImage 2.1.4 (was 3.0.0)

This gives 4 build errors - screenshots below
Screenshot 2020-09-08 at 20 50 41
Screenshot 2020-09-08 at 20 50 52
Screenshot 2020-09-08 at 20 51 00
Screenshot 2020-09-08 at 20 51 10

Hope this helps in recreating the issue.

sagesse-cn:
> Specifically, if you could point out how I can reproduce it, that would be very helpful!


#### Reproduce

In any dependent PINCache project.

```
PINMemoryCache *cache = [[PINMemoryCache alloc] init];
cache.didAddObjectBlock = ^(PINMemoryCache *cache, NSString * _Nonnull key, id  _Nullable object) {
 };
```

### Reason

The compile error in PINCache project was fixed in #276.
 
But that solution was not the best, any users for using PINCache may need to make a lot of changes.

A better solution is to add the `__kindof` into `id<PINCaching>`, user not need to modify code, just simply update the Cocoapods dependency.

About why must to add `__kindof`,  see reference.

Thx!

Plnda:
Could we get some traction on this PR, What is currently blocking it?

natebirkholz:
What is holding up approval of this PR? I cannot get my work done without it.

garrettmoon (approved):
Thank you for the explanation and fix! Hopefully this will make downstream dependencies a little easier to manage.

natebirkholz:
> Reference: https://reviews.llvm.org/D66831#2019125

I get the latest with your changes and I still get two incompatible pointers.

PINCache.m line 109;
` [self->_memoryCache objectForKeyAsync:key completion:^(PINMemoryCache`
and line 232:
`[_diskCache synchronouslyLockFileAccessWhileExecutingBlock:^(PINDiskCache *diskCache)`

natebirkholz:
> Reference: https://reviews.llvm.org/D66831#2019125

@sagesse-cn

sagesse-cn:
> > Reference: https://reviews.llvm.org/D66831#2019125
> 
> @sagesse-cn

Can you provide more information, a demo that depyon to pincache?

natebirkholz:
> > > Reference: https://reviews.llvm.org/D66831#2019125
> > 
> > 
> > @sagesse-cn
> 
> Can you provide more information, a demo that depyon to pincache?

@sagesse-cn My bad, the pod updates but something somewhere in this large legacy project I took over is specifying an earlier version of PINCache, so in my Xcode project it is an old version.

PR #283, opened 2020-09-14
Added SPM support

3a4oT:
Added SPM support starting from Swift 5.3 (since it supports resources). Some notes:

- SPM tests not passing when you run it from a command line (SPM(bug)? gets the wrong bundle path) but passing if you use Xcode in SPM edit mode(just open as a package instead of .xcodeproj).

- dependency strategy for `PINOperation` set as - branch `master`. I believe it should be set - from `PINOperation 1.2.0` but unfortunately, it was tagged as `1.2` and SPM can not resolve it. Can the maintainers re-tag/new it with proper sem-ver?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -52,10 +52,10 @@ jobs:
   SPM-Build:
     runs-on: macos-latest
     env:
-      DEVELOPER_DIR: /Applications/Xcode_12_beta.app/Contents/Developer
+      DEVELOPER_DIR: /Applications/Xcode_12.app/Contents/Developer
     steps:
       - name: Checkout
         uses: actions/checkout@v2
 
       - name: Verify that PINCache can be build by SPM
-        run: xcrun swift build
+        run: make spm
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -37,15 +37,25 @@ jobs:
     name: CocoaPods
     runs-on: macOS-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Lint
         run: pod lib lint
   carthage:
     name: Carthage
     runs-on: macOS-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Update
         run: carthage update --no-use-binaries --no-build
       - name: Build
         run: carthage build --no-skip-current
+  SPM-Build:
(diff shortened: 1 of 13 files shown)


Tests/PINDiskCache+PINCacheTests.m
```diff
@@ -15,7 +15,7 @@ @implementation PINDiskCache (PINCacheTests)
 - (void)setTtlCacheSync:(BOOL)ttlCache
 {
     [self lock];
-        self->_ttlCache = ttlCache;
+    [self setValue:@(ttlCache) forKey:@"_ttlCache"];
```
3a4oT, line 18: Fixed build issue when testing package in Xcode.

xezero:
This would be fantastic if this got merged! :)

3a4oT:
@garrettmoon ping...

.github/workflows/ci.yaml
```diff
@@ -37,15 +37,25 @@ jobs:
...
+  SPM-Build:
+    runs-on: macos-latest
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_12_beta.app/Contents/Developer
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Verify that PINCache can be build by SPM
+        run: xcrun swift build
```
garrettmoon, line 61: Do you mind moving this to the Makefile and calling `make spm` here?

.github/workflows/ci.yaml
```diff
@@ -37,15 +37,25 @@ jobs:
...
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Update
         run: carthage update --no-use-binaries --no-build
       - name: Build
         run: carthage build --no-skip-current
+  SPM-Build:
+    runs-on: macos-latest
+    env:
+      DEVELOPER_DIR: /Applications/Xcode_12_beta.app/Contents/Developer
```
garrettmoon, line 55: Can this be updated now that 12 is released?

garrettmoon (commented):
Do you mind adding a command to the make file as well

3a4oT:
@garrettmoon, I've addressed your feedback. Thanks!

3a4oT:
@garrettmoon  This seems to be a sporadic failure, I can confirm that from time to time the same test fails locally. Can we trigger tests again?

garrettmoon (approved):
Looks good to me! Thanks so much for adding the support for SPM!

PR #296, opened 2020-11-05
Fix a race between -[removeAllObjects] and -[removeObjectForKey:] which could cause an item to not be removed

### LINKED ISSUE #295: PINDiskCache: +[moveItemAtURLToTrash:] races with +[emptyTrash]
Proof of concept:

```
- (void)testDiskEmptyTrashRace
{
    NSString *cacheName = @"testDiskEmptyTrashRace";
    PINDiskCache *diskCache = [[PINDiskCache alloc] initWithName:cacheName];

    for (NSUInteger idx = 0; idx < 10000; idx++) {
        NSString *key = [@(idx) stringValue];
        NSData *tmpData = [key dataUsingEncoding:NSUTF8StringEncoding];
        [diskCache removeAllObjects];
        [diskCache setObject:tmpData forKey:key];
        XCTAssertNotNil([diskCache objectForKey:key]);
        [diskCache removeObjectForKey:key];
        XCTAssertNil([diskCache objectForKey:key]);
    }
}
```

The `XCTAssertNil` check fails eventually:

Test failure

`-[PINDiskCache removeAllObjects]` calls `+[PINDiskCache emptyTrash]`. This calls `dispatch_async(...)`, which schedules the block but doesn't execute it.

`-[PINDiskCache removeObjectForKey:]` (indirectly) calls `+[PINDiskCache moveItemAtURLToTrash:]`. This call is protected by `_mutex`, but not by `sharedLock`. This method then calls `+[PINDiskCache sharedTrashURL]` to obtain the shared trash URL.

However, in between the time when the URL is retrieved and the item is moved to the trash URL, the async block from `+[PINDiskCache emptyTrash]` executes. This block removes the shared trash directory (it also clears `_sharedTrashURL` but that's irrelevant because `+[PINDiskCache moveItemAtURLToTrash:]` has already obtained the old trash URL).

Now that the trash directory is deleted, `+[PINDiskCache moveItemAtURLToTrash:]` calls `-[NSFileManager moveItemAtURL:toURL:error:]` to attempt to move the item into the trash. This will fail because the trash directory no longer exists. The end result is that the item is never removed. This is a time-of-check-to-time-of-use race.

There are a couple ways this could be fixed.

- Option A: If the call to `+[PINDiskCache moveItemAtURLToTrash:]` fails, the cache could have a fallback and just directly call `-[NSFileManager removeItemAtURL:error:]` to delete the item.
- Option B: `+[PINDiskCache moveItemAtURLToTrash:]` could obtain the `sharedLock` and hold it while executing the following two lines of code:
    ```
    NSURL *uniqueTrashURL = [[PINDiskCache sharedTrashURL] URLByAppendingPathComponent:uniqueString isDirectory:NO];
    BOOL moved = [[NSFileManager defaultManager] moveItemAtURL:itemURL toURL:uniqueTrashURL error:&error];
    ```
    This requires some refactoring to either use a recursive lock or make it so the call to `+[sharedTrashURL]` doesn't deadlock.

Option A is probably the simplest, but Option B probably has less mental overhead because it fully resolves the race condition rather than trying to recover from it ad-hoc.

Personally, I like both options. I like that Option B fully resolves the race, and I like that Option A provides a fallback. If you're trying to delete a file, I don't think you should give up and keep the file just because you failed to create a new directory or rename something, since those actions are orthogonal to the original goal.

mjbshaw:
This fixed #295

### DIFF
--- a/Source/PINDiskCache.m
+++ b/Source/PINDiskCache.m
@@ -424,37 +424,43 @@ static NSURL *_sharedTrashURL;
 
 + (NSURL *)sharedTrashURL
 {
-    NSURL *trashURL = nil;
-    
-    [[PINDiskCache sharedLock] lock];
-        if (_sharedTrashURL == nil) {
-            NSString *uniqueString = [[NSProcessInfo processInfo] globallyUniqueString];
-            _sharedTrashURL = [[[NSURL alloc] initFileURLWithPath:NSTemporaryDirectory()] URLByAppendingPathComponent:uniqueString isDirectory:YES];
-            
-            NSError *error = nil;
-            [[NSFileManager defaultManager] createDirectoryAtURL:_sharedTrashURL
-                                     withIntermediateDirectories:YES
-                                                      attributes:nil
-                                                           error:&error];
-            PINDiskCacheError(error);
-        }
-        trashURL = _sharedTrashURL;
-    [[PINDiskCache sharedLock] unlock];
-    
-    return trashURL;
+    NSAssert(![[PINDiskCache sharedLock] tryLock] || ([[PINDiskCache sharedLock] unlock], NO),
+             @"+[sharedTrashURL] must be called with +[sharedLock] held.");
+    if (_sharedTrashURL == nil) {
+        NSString *uniqueString = [[NSProcessInfo processInfo] globallyUniqueString];
+        _sharedTrashURL = [[[NSURL alloc] initFileURLWithPath:NSTemporaryDirectory()] URLByAppendingPathComponent:uniqueString isDirectory:YES];
+
+        NSError *error = nil;
+        [[NSFileManager defaultManager] createDirectoryAtURL:_sharedTrashURL
+                                 withIntermediateDirectories:YES
+                                                  attributes:nil
+                                                       error:&error];
+        PINDiskCacheError(error);
+    }
+    return _sharedTrashURL;
 }
 
-+ (BOOL)moveItemAtURLToTrash:(NSURL *)itemURL
++ (BOOL)moveItemAtURLToTrashOrRemove:(NSURL *)itemURL
 {
     if (![[NSFileManager defaultManager] fileExistsAtPath:[itemURL path]])
         return NO;
     
     NSError *error = nil;
     NSString *uniqueString = [[NSProcessInfo processInfo] globallyUniqueString];
-    NSURL *uniqueTrashURL = [[PINDiskCache sharedTrashURL] URLByAppendingPathComponent:uniqueString isDirectory:NO];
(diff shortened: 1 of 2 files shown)


garrettmoon:
Great writeup in the issue and fix, thank you! Would you mind checking to see if the Carthage error is related to your change? Seems to be tvOS specific?

mjbshaw:
I don't think the Carthage error is related. Running `make carthage` locally on master also fails:

```
$ git clone https://github.com/pinterest/PINCache.git
Cloning into 'PINCache'...
remote: Enumerating objects: 42, done.
remote: Counting objects: 100% (42/42), done.
remote: Compressing objects: 100% (30/30), done.
remote: Total 3811 (delta 12), reused 24 (delta 7), pack-reused 3769
Receiving objects: 100% (3811/3811), 13.96 MiB | 11.82 MiB/s, done.
Resolving deltas: 100% (2142/2142), done.
$ cd PINCache
$ make carthage
carthage update --no-use-binaries --no-build
*** Fetching PINOperation
*** Checking out PINOperation at "1.2.1"
carthage build --no-skip-current
*** xcodebuild output can be found in /var/folders/d_/hj61jd596lx70cd65k9f8gk8009nf7/T/carthage-xcodebuild.3xheOw.log
*** Building scheme "PINOperation-macOS" in PINOperation.xcodeproj
*** Building scheme "PINOperation-watchOS" in PINOperation.xcodeproj
*** Building scheme "PINOperation-tvOS" in PINOperation.xcodeproj
Build Failed
	Task failed with exit code 1:
	/usr/bin/xcrun lipo -create /Users/mjbshaw/Library/Caches/org.carthage.CarthageKit/DerivedData/12.1_12A7403/PINOperation/1.2.1/Build/Intermediates.noindex/ArchiveIntermediates/PINOperation-tvOS/IntermediateBuildFilesPath/UninstalledProducts/appletvos/PINOperation.framework/PINOperation /Users/mjbshaw/Library/Caches/org.carthage.CarthageKit/DerivedData/12.1_12A7403/PINOperation/1.2.1/Build/Products/Release-appletvsimulator/PINOperation.framework/PINOperation -output /tmp/PINCache/Carthage/Build/tvOS/PINOperation.framework/PINOperation

This usually indicates that project itself failed to compile. Please check the xcodebuild log for more details: /var/folders/d_/hj61jd596lx70cd65k9f8gk8009nf7/T/carthage-xcodebuild.3xheOw.log
make: *** [carthage] Error 1
```

garrettmoon:
Ah, indeed you are correct. I've fixed it in master. Do you mind rebasing?

mjbshaw:
Done. All checks pass now. Thanks for fixing that!

garrettmoon (approved):
Thanks again for the fix!

PR #301, opened 2020-11-17
Fixed spm integration on regular Xcode project.

3a4oT:
- without this change project itself can be built as spm package, but failed to do so when you integrate it as a dependency via Xcode.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,11 @@ PLATFORM="platform=iOS Simulator,name=iPhone 8"
 SDK="iphonesimulator"
 SHELL=/bin/bash -o pipefail
 XCODE_MAJOR_VERSION=$(shell xcodebuild -version | HEAD -n 1 | sed -E 's/Xcode ([0-9]+).*/\1/')
+IOS_EXAMPLE_PROJECT="IntegrationTests/PINCache-SPM-Integration/PINCache-SPM-Integration.xcodeproj"
+EXAMPLE_PROJECT_CARTHAGE_WORKAROUND="IntegrationTests/PINCache-SPM-Integration/PINCache-SPM-Integration.carthageWorkaround"
+EXAMPLE_SCHEME="PINCache-SPM-Integration"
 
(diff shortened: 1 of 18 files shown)


garrettmoon:
Thanks for putting this up! I hate to ask, but would you mind updating the CI test to ensure that SPM works when integrating into a project?

3a4oT:
Removing `shared` from the scheme doesn't work because of Xcode/Swift versions. 

```
xcodebuild: error: Could not resolve package dependencies:
43
  package at '/Users/runner' is using Swift tools version 5.3.0 but the installed version is 5.2.0
```

3a4oT:
Have applied the workaround script which allows us to use Xcode 12. If the concept is acceptable I will do a similar thing for PINRemoteImage.

Makefile
```diff
@@ -2,8 +2,11 @@ PLATFORM="platform=iOS Simulator,name=iPhone 8"
 SDK="iphonesimulator"
 SHELL=/bin/bash -o pipefail
 XCODE_MAJOR_VERSION=$(shell xcodebuild -version | HEAD -n 1 | sed -E 's/Xcode ([0-9]+).*/\1/')
+IOS_EXAMPLE_PROJECT="IntegrationTests/PINCache-SPM-Integration/PINCache-SPM-Integration.xcodeproj"
+EXAMPLE_PROJECT_CARTHAGE_WORKAROUND="IntegrationTests/PINCache-SPM-Integration/PINCache-SPM-Integration.carthageWorkaround"
```
garrettmoon, line 6: Is EXAMPLE_PROJECT_CARTHAGE_WORKAROUND used anywhere?

garrettmoon (approved):
Looks good barring a minor nit! I'll give this a local test tomorrow and merge if everything looks good. Thanks so much for your contribution!

Makefile
3a4oT: good catch! should be fixed now!

PR #312, opened 2021-11-02
[Fix] Disk cache always trims cache to default size on creation

tjwio:
There's a bug currently where `PINDiskCache` will always trim cache size to 50MB when creating since it defaults to 50MB in the init, and immediately trims cache in `[self initializeDiskProperties];`. This is obviously not ideal since if we specify a cache size >50MB then we will constantly be having to refetch assets on each app open since it will trim to 50MB each time.

Proposed fix:
- Add two new fields to designated init to allow passing in custom `byteLimit` and `ageLimit`
- Keep convenience init to maintain API stability that defaults to 50MB/30 days


garrettmoon (approved):
This is a good change since it makes PINDiskCache less tricksy, thanks for the fix!

Some more context: This shouldn't happen in practice if the cache limit is set immediately after init since the trim is triggered on a separate queue and only checks the limit after doing a bunch of disk IO. So while I'm skeptical this is affecting many library users, I think this makes things *much* clearer.

PR #315, opened 2022-10-25
Only sort accessDates if totalCost exceeds costLimit

kchronis:
As the number of cached objects increases, sorting the accessDates becomes expensive. The access dates should only be sorted if the totalCost is greater than the limit.


garrettmoon (approved):
Looks like a good change to me!

PR #317, opened 2022-10-26
Don't hold lock while sorting

garrettmoon:
Based on what @kchronis discovered, this should help prevent lock contention further.


kchronis:
LGTM

PR #320, opened 2023-04-08
Fix Xcode 14.3 which supports only iOS 11.0+

kenstir:


kenstir:
This was superseded by #327 .  Closing.

Issue #324, opened 2024-02-01
Privacy manifest


lamdaniel:
Apple's upcoming privacy manifest changes https://developer.apple.com/news/?id=r1henawx

From my understand for required reason API's https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

I believe we use 
- NSFileModificationDate
- NSURLCreationDateKey
- NSURLContentModificationDateKey

Will there be a privacy manifest added to include these?

Lax:
I got another notification that requires to add this key

* NSPrivacyAccessedAPICategoryFileTimestamp

PR #326, opened 2024-04-29
Add PrivacyInfo.xcprivacy to podspec

gazadge:
Related to https://github.com/pinterest/PINCache/issues/324.

Added `PrivacyInfo.xcprivacy` to `PINCache.podspec` so it's included in Resources when installing via Cocoapods.


jparise:
The change looks straightforward, but I think we should understand the CI build failures before merging.

PR #328, opened 2024-05-02
Add Least Frequently Used eviction strategy

andyfinnell:
## Summary

Currently PINCache only offers LRU (least recently used) as an eviction strategy. However, there are some special workloads where LFU (least frequently used) could offer better performance. This PR introduces LFU alongside the existing LRU eviction strategy. The default is still LRU.

There is also some minor renaming to the `trimToSizeByDateAsync`, `trimToSizeByDate`, `trimToCostByDate`, and `trimToCostByDateAsync` methods, since those now follow the explicit eviction strategy. Old methods remain and work as expected, but are marked deprecated.

## Testing

Added some unit tests for both memory and disk caches to verify objects are evicted based on access count when LFU is selected. Ran tests on iOS, tvOS, macOS.


Source/PINCache.m
```diff
@@ -76,8 +88,11 @@ - (instancetype)initWithName:(NSString *)name
                                              keyEncoder:keyEncoder
                                              keyDecoder:keyDecoder
                                          operationQueue:_operationQueue
-                                               ttlCache:ttlCache];
-        _memoryCache = [[PINMemoryCache alloc] initWithName:_name operationQueue:_operationQueue ttlCache:ttlCache];
+                                               ttlCache:ttlCache
+                                              byteLimit:50 * 1024 * 1024 // 50 MB by default
+                                               ageLimit:60 * 60 * 24 * 30 // 30 days by default
```
rcancro, line 93: Should these be constants?

Source/PINDiskCache.m
```diff
@@ -201,6 +204,33 @@ - (instancetype)initWithName:(NSString *)name
                      ageLimit:60 * 60 * 24 * 30]; // 30 days by default
```
rcancro, line 204: Oh I see these were never constants! Now that we are using them in two places, let's create constants that we can use in both places.

rcancro (approved):

Source/PINCache.m
andyfinnell: Done!

PR #336, opened 2026-07-29
Fix crashes and unbounded memory in disk cache trimming

dreed1:
trimToSize:, trimToSizeByEvictionStrategy:, and trimToDate: sorted the entire metadata dictionary with keysSortedByValueUsingComparator:, which materializes O(N) temporary storage inside CoreFoundation. On large caches under memory pressure that allocation fails and traps in __CFCreateArrayStorage — and since every subsequent trim hits the same wall, a cache that grows too big to sort can never shrink again.

- Select eviction candidates in bounded 256-key batches (single pass, binary-search insertion, O(batch) memory) and loop until under the byte target, so trims always make progress. An attempted-keys set guarantees termination when eviction declines ghost entries.
- trimToDate: is a pure threshold filter; drop its sort entirely.
- Nil-safe comparators: metadata dates/size are nil when resourceValuesForKeys: fails during the initial disk scan. [NSNumber compare:nil] RAISES NSInvalidArgumentException (a real crash in trimToSize:), and [NSDate compare:nil] returns NSOrderedSame, which made the sorts weakly inconsistent. Unknown-metadata entries now sort first and are evicted before entries with known access times.
- Guard NSUInteger underflow in the trim loops: stale metadata sizes overshooting byteCount made the break condition unsatisfiable and evicted the entire cache.

Adds a multi-batch trim regression test (300 entries > one batch).


Source/PINDiskCache.m
```diff
@@ -66,6 +105,48 @@ @interface PINDiskCacheMetadata : NSObject
...
+    NSDictionary<NSString *, PINDiskCacheMetadata *> *metadata,
+    NSUInteger limit,
+    NSSet<NSString *> *excludedKeys,
+    PINDiskCacheEvictionPriorityComparator priorityComparator)
+{
+    NSMutableArray<NSString *> *candidates = [[NSMutableArray alloc] initWithCapacity:limit + 1];
+    NSComparisonResult (^keyComparator)(NSString *, NSString *) = ^NSComparisonResult(NSString *key1, NSString *key2) {
+        return priorityComparator(metadata[key1], metadata[key2]);
+    };
+    [metadata enumerateKeysAndObjectsUsingBlock:^(NSString *key, PINDiskCacheMetadata *entryMetadata, BOOL *stop) {
```
andyfinnell, line 129: Is this not an insertion sort? i.e. O(n^2)? Or am I missing something?

Source/PINDiskCache.m
```diff
@@ -54,6 +54,45 @@ typedef NS_ENUM(NSUInteger, PINDiskCacheCondition) {
     return url.fileSystemRepresentation;
 }
 
+// Metadata dates and size are nil when -resourceValuesForKeys: failed during the
+// initial disk scan (e.g. the file disappeared mid-scan). Passing nil to -compare:
```
andyfinnell, line 58: If the date and size are nil, does that always mean the file has disappeared? i.e. should we always evict those missing the date/size?

andyfinnell (approved):
It'd be nice if we could put the cap behind an experiment to determine what performance impact it has, but probably not worth the effort.

Source/PINDiskCache.m
dreed1: it does have the shape of one, but the sorted buffer is capped at 256 and the cap makes it okay in my mind?

Once the buffer is full, we start to check this against the current worst candidate and it gets rejected in O(1) if it doesn't beat it. If it *does* qualify though, we do a binary search for the index (which is something like O(log256)), then we drop the last element after insertion.

Source/PINDiskCache.m
dreed1: not always -- with NSFileProtectionComplete-class files while the device is locked, attribute reads can fail even though the file is intact and perfectly readable  once unlocked. if we evict regardless we could be wrong there.

